PR #7, opened 2016-11-28
[DO NOT MERGE] Added a new Ansible role to support change hostname for a Satellite host

sthirugn:
This is a first rough draft submission.  This needs some polishing before merging.


roles/change_hostname/files/change_hostname.rb
```diff
@@ -0,0 +1,83 @@
...
+#
+# 1) log into UI, rename the default capsule under infrastructure->capsules to the new hostname in both the name and url fields
+#
+# 2) update NEW_HOSTNAME variable below
+#
+require "socket"
+
+STDOUT.puts "Starting the script run..."
+
+# update this variable
```
johnpmitsch: You can remove the "update this variable" note

roles/change_hostname/files/change_hostname.rb
```diff
@@ -0,0 +1,83 @@
...
+NEW_HOSTNAME = ARGV[0]
+
+# don't change this variable, it is getting the hostname from your system
```
johnpmitsch: Same here

roles/change_hostname/files/change_hostname.rb
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/ruby
+
+# BEFORE RUNNING THIS SCRIPT
+#
+# 1) log into UI, rename the default capsule under infrastructure->capsules to the new hostname in both the name and url fields
+#
+# 2) update NEW_HOSTNAME variable below
```
johnpmitsch: Also can remove this step since its a command line argument

roles/change_hostname/files/change_hostname.rb
```diff
@@ -0,0 +1,83 @@
...
+#
+require "socket"
+
+STDOUT.puts "Starting the script run..."
+
+# update this variable
+NEW_HOSTNAME = ARGV[0]
```
johnpmitsch: Can we add a check for more than one argument or no arguments that prints out something like
```USAGE: ./change_hostname.rb <new-hostame>```?

Comment (sthirugn) [inline roles/change_hostname/files/change_hostname.rb]: This script is currently called from inside the ansible role  `command: /tmp/change_hostname.rb {{ NEW_HOSTNAME }}`  and will not be changed by the user hopefully.  Do we still need this check?

File: roles/change_hostname/files/change_hostname.rb (hunk 2 above)
Comment (sthirugn) [inline roles/change_hostname/files/change_hostname.rb]: Updated

File: roles/change_hostname/files/change_hostname.rb (hunk 1 above)
Comment (sthirugn) [inline roles/change_hostname/files/change_hostname.rb]: Updated

File: roles/change_hostname/files/change_hostname.rb (hunk 3 above)
Comment (sthirugn) [inline roles/change_hostname/files/change_hostname.rb]: Updated

File: roles/change_hostname/files/change_hostname.rb [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+`
+
+STDOUT.puts "updating hostname in /etc/hostname"
+`sed -i -e 's/#{OLD_HOSTNAME}/#{NEW_HOSTNAME}/g' /etc/hostname`
+
+STDOUT.puts "updating hostname in /etc/hosts"
+`sed -i -e 's/#{OLD_HOSTNAME}/#{NEW_HOSTNAME}/g' /etc/hosts`
+
+STDOUT.puts "setting hostname"
+`hostnamectl set-hostname #{NEW_HOSTNAME}`
```
pgagne: For changing the hostname for the system, ie hostnamectl or editing /etc/sysconfig/network, do we want to leave this up to ansible and use there builtin module to allow better RHEL version independence?

johnpmitsch: @pgagne is right, we still have to change this - @sthirugn I suggest we check for version and then update the hostname differently if its rhel6 or 7, I can write something when I have a minute

sthirugn:
The commit in this PR is merged with https://github.com/RedHatSatellite/satellite-clone/pull/9.  Closing this PR.

PR #28, opened 2016-12-21
Remove unused sat6repro roles files

sthirugn:
The removed files are necessary only when we start pushing roles to ansible galaxy.  These can be added later when needed.


sthirugn:
rebased

sthirugn:
Rebased

PR #30, opened 2016-12-21
Replace `inventory` with `inventory.sample`

sthirugn:
This is done to:
1. Facilitate easy contribution back to upstream.
2. Permanently save some hosts/ip addresses locally for future reference.


johnpmitsch:
good change :+1:

Issue #35, opened 2017-01-03
libselinux-python package install step should be after when instance is subscribed to rhel repos


sghai:
package install [step](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/sat6repro/tasks/main.yml#L30) should be after [this](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/sat6repro/tasks/main.yml#L30)
 otherwise, install may failed with error:

```console
 FAILED! => {"changed": false, "failed": true, "msg": "There are no enabled repos.\n Run \"yum repolist all\" to see the repos you have.\n You can enable repos with yum-config-manager --enable <repo>\n", "rc": 1, "results": []}
```

sthirugn:
PR #36 is merged now.

PR #36, opened 2017-01-03
Installing libselinux-python package after enabling base os repo

sghai:
handles #35


sthirugn:
Good catch, thank you for the PR.

PR #47, opened 2017-01-12
Changed URL in 'git clone' to HTTPS.

pgagne:
HTTPS is better for anon access, since you don't need to setup keys.


sthirugn:
I agree https is better for anon access and git is better for contributing back to upstream.

Issue #46, opened 2017-01-10
After starting katello services, hammer ping fails as foreman_tasks doesn't comes up immediately


sghai:
```console
TASK [sat6repro : Restart katello-service and test satellite] ******************
changed: [Server] => (item=katello-service start)
failed: [Server] (item=hammer ping) => {"changed": true, "cmd": ["hammer", "ping"], "delta": "0:00:54.061257", "end": "2017-01-10 04:35:41.361823", "failed": true, "item": "hammer ping", "rc": 1, "start": "2017-01-10 04:34:47.300566", "stderr": "", "stdout": "candlepin:      \n    Status:          ok\n    Server Response: Duration: 35ms\ncandlepin_auth: \n    Status:          ok\n    Server Response: Duration: 36ms\npulp:           \n    Status:          ok\n    Server Response: Duration: 63ms\nforeman_tasks:  \n    Status:          FAIL\n    Server Response:", "stdout_lines": ["candlepin:      ", "    Status:          ok", "    Server Response: Duration: 35ms", "candlepin_auth: ", "    Status:          ok", "    Server Response: Duration: 36ms", "pulp:           ", "    Status:          ok", "    Server Response: Duration: 63ms", "foreman_tasks:  ", "    Status:          FAIL", "    Server Response:"], "warnings": []}
```

This appears only when we use a large db and it might be because when multiple foreman_tasks are running in backend

sghai:
I manually checked at server the output of hammer ping and got same issue, but after a few minutes it came up successfully.

```
~]# hammer -u admin -p changeme ping
candlepin:      
    Status:          ok
    Server Response: Duration: 27ms
candlepin_auth: 
    Status:          ok
    Server Response: Duration: 27ms
pulp:           
    Status:          ok
    Server Response: Duration: 27ms
foreman_tasks:  
    Status:          FAIL
    Server Response:

~]# hammer -u admin -p changeme ping
candlepin:      
    Status:          ok
    Server Response: Duration: 26ms
candlepin_auth: 
    Status:          ok
    Server Response: Duration: 27ms
pulp:           
    Status:          ok
    Server Response: Duration: 26ms
foreman_tasks:  
    Status:          ok
    Server Response: Duration: 1441ms
```

See the duration taken by `foreman_tasks`. This needs to be handled.

sghai:
The ugly solution that I can think of putting a sleep here
[...]

PR #48, opened 2017-01-13
Added a sleep before hammer ping so that foreman-task service cleanly comes up

sghai:
handles #46

### DIFF
--- a/roles/sat6repro/tasks/main.yml
+++ b/roles/sat6repro/tasks/main.yml
@@ -91,11 +91,14 @@
 - include: restore.yml
   when: perform_restore
 
-- name: Restart katello-service and test satellite
-  command: "{{ item }}"
-  with_items:
-    - katello-service start
-    - hammer ping
+- name: Restart katello-service
+  command: katello-service start
+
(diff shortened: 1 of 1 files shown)


sthirugn:
As per Sachin, adding 3 minutes didnt solve the problem, so he added 5 minutes.

sthirugn:
ACK

roles/sat6repro/tasks/main.yml
```diff
@@ -91,11 +91,14 @@
...
-- name: Restart katello-service and test satellite
-  command: "{{ item }}"
-  with_items:
-    - katello-service start
-    - hammer ping
+- name: Restart katello-service
+  command: katello-service start
+
+- name: Wait for foreman-tasks service to comes up
+  shell: sleep 5m
```
sghai: @sthirugn : I was reading about ansible before making change and realized that pause is for playbook but not for tasks. But as you suggested 'pause' looks more promising. will update the PR.

sthirugn:
Thanks @sghai

PR #49, opened 2017-01-13
Remove rsync install task on the Ansible Control node.

### LINKED ISSUE #37: rsync installation errors in Ansible Control node
Failed in Fedora 25:
Two issues:
1. The task assumes that the user needs to be a root in Ansible control node - which is not desirable.
2. rsync installation failure:
TASK [sat6repro : Install rsync on the Ansible Control Node] *******************
fatal: [10.8.0.73 -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failure talking to yum: failure: repodata/repomd.xml from fedora-cisco-openh264: [Errno 256] No more mirrors to try.\nhttps://codecs.fedoraproject.org/openh264/25/x86_64/repodata/repomd.xml: [Errno -1] repomd.xml signature could not be verified for fedora-cisco-openh264"}

I think the first issue is more serious than the second.

sthirugn:
Reasons:
1. The tasks mentioned in the sat6repro/tasks/main.yml should be focusing only on updating the Satellite host
2. Installing of a task requires that Ansible needs to run as a root user on the Control node which is not a recommended practice.

Fixes #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,17 @@ Ideally, you need two hosts to run this project:
 - RHEL 6
 - RHEL 7
 
-1. Install `ansible` package on the Control node. For RHEL boxes, [access to EPEL] (https://access.redhat.com/solutions/3358) is required.
+1. Installation of required packages:
+   a. Install `ansible` package on the Control node. For RHEL boxes, [access to EPEL] (https://access.redhat.com/solutions/3358) is required.
 
-  ```console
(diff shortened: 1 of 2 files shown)


sthirugn:
I waited for two days to get @johnpmitsch's comments here but looks like he is out.  I am going to merge this now.  Any concerns can be created as new issues.

PR #52, opened 2017-01-17
Improvements to free space calculation.

sthirugn:
1. Replace df with ansible facts to calculate the free space of the host
2. Introducing a new task file to hold the pre-install checks
3. Giving an option to the user to disable pre-install checks if required

### DIFF
--- a/roles/sat6repro/tasks/main.yml
+++ b/roles/sat6repro/tasks/main.yml
@@ -1,14 +1,9 @@
 ---
 - include_vars: main.yml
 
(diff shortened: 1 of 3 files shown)


sthirugn:
I tested this on both rhel6 and rhel7 boxes.

Use cases tested:
1. if freespace < 75 GB
2. if freespace > 75 GB
3. run_pre_install_check = true
4. run_pre_install_check = false

sthirugn:
Great! thanks.

PR #54, opened 2017-01-18
Require libselinux-python in rhel6 sooner

### LINKED ISSUE #53: libselinux-python required for RHEL6
Ran the clone-tool with the destination as the control host. Running RHEL6.

Looks like we need libselinux-python on the destination host maybe?
```
TASK [sat6repro : create /etc/hosts] *******************************************
fatal: [10.8.0.146]: FAILED! => {"changed": true, "failed": true, "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}
	to retry, use: --limit @/root/suresh-sat-clone/satellite-clone-playbook.retry
```

sthirugn:
Fixes #53

### DIFF
--- a/roles/sat6repro/tasks/main.yml
+++ b/roles/sat6repro/tasks/main.yml
@@ -5,6 +5,10 @@
 - include: pre_install_check.yml
   when: run_pre_install_check
 
+# Install libselinux package
+- name: install libselinux-python for rhel6
+  yum: name=libselinux-python state=latest
+
 # turn off firewall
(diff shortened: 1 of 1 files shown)


sthirugn:
<pgagne> sthirugn dont we only want to install libselinux-python on rhel6. your code does it on both
<sthirugn> pgagne: rhel7 already comes with it, if iirc
<sthirugn> pgagne: I left it just to ensure that rhel7 has it

PR #56, opened 2017-01-18
Introducing control node setup script to ease the process for new users

### LINKED ISSUE #55: Create a bash script to ease setting up of Ansible control node

sthirugn:
This script will do the following for the user on the Ansible control node:
1. Register the node to CDN
2. Install epel
3. Install ansible
4. Install required packages for Ansible modules
5. git clone satellite-clone
6. Copies required config files in required locations

As explained above, this automates most of the manual steps in Ansible control node.  After running this script, the user just needs to plug in data in the created config files.

Support: The script supports RHEL6, RHEL7 and Fedora.

Fixes #55


johnpmitsch:
@sthirugn Nice script, what is your idea of a workflow for a user? If they clone this repo to run the script, this will re-clone the repo

sthirugn:
@johnpmitsch I was thinking about asking them to run the script directly from github if they want to avoid setting up control node manually.

sthirugn:
Workflow:
1. Run the setup script
2. Update the configs (which are already created by the script)
3. Run the playbook
4. Go for coffee break!

johnpmitsch:
above discussion resolved on IRC ^

@sthirugn the variables would be nice if they could be passed as named arguments. ex `./setup.sh --username foo --password bar --rhel-major-version 7` and checked to see if they exist

sthirugn:
@johnpmitsch if you are already registered with sub manager, you will get a warning and that step would be skipped

bash named arguments - I can create a new issue to track one this pr is merged.

README.md
```diff
@@ -20,7 +20,19 @@ Ideally, you need two hosts to run this project:
...
+1. git clone this project.
+
+  ```console
+     # git clone https://github.com/RedHatSatellite/satellite-clone.git
+  ```
+  NOTE: Optionally now run satellite-clone/helpers/control_node_setup.sh which will setup required packages and files for you.
+
+  ```console
+    # cd satellite-clone
+    # /bin/bash helpers/control_node_setup.sh
```
johnpmitsch: wouldn't the users have to update the variables in the setup script?

sthirugn:
@johnpmitsch updated

helpers/control_node_setup.sh
```diff
@@ -0,0 +1,38 @@
...
+# RHEL7: 7
+# Fedora: Comment or remove the below line
+export RHEL_MAJOR_VERSION=7
+
+if [ -z "$RHEL_MAJOR_VERSION" ]; then
+  dnf install -y ansible
+else
+  # Register and subscribe
+  echo "Register and subscribe the node. This is done to install few packages."
+  subscription-manager register --username="$USERNAME" --password="$PASSWORD" --autosubscribe
```
johnpmitsch: It would be nice to add a check for either:
- the `$USERNAME`  and `$PASSWORD` variables are updated
- this command completes successfully

and terminate the script if this isn't true

helpers/control_node_setup.sh
```diff
@@ -0,0 +1,38 @@
...
+
+  set -x
+  # Enable epel repo and install Ansible
+  wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-"$RHEL_MAJOR_VERSION".noarch.rpm
+  yum localinstall epel-release-latest-"$RHEL_MAJOR_VERSION".noarch.rpm
```
pgagne: Is there a reason you just don't do:

```
yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-"$RHEL_MAJOR_VERSION".noarch.rpm
```

This will work in RHEL6 and RHEL7

helpers/control_node_setup.sh
sthirugn: Added your enhancement suggestion to https://github.com/RedHatSatellite/satellite-clone/issues/57.

johnpmitsch: @sthirugn could we at least check that the username and password variables are updated here? I could see many users running the script without updating it and this would alert them to that

johnpmitsch: This should work:

`if [[ $USERNAME -eq "changeme" && $PASSWORD -eq "changeme" ]] ; then  echo "Please update variables in the script";   exit 1; fi` (you will have to change it from a one-liner)

sthirugn: Thanks! Right fix is:
```sh
if [[ $USERNAME = "changeme" && $PASSWORD = "changeme" ]]; then
  echo "Please update the variables in the script before running it."
  exit 1
fi
```

sthirugn:
Updated as per review comments and also improved README.

helpers/control_node_setup.sh
```diff
@@ -0,0 +1,40 @@
...
+if [[ $USERNAME = "changeme" && $PASSWORD = "changeme" ]]; then
+  echo "Please update the variables in the script before running it."
+  exit 1
+fi
+
+# RHEL_MAJOR_VERSION of the Control node should be as follows:
+# RHEL6: 6
+# RHEL7: 7
+# Fedora: Comment or remove the below line
+export RHEL_MAJOR_VERSION=7
```
johnpmitsch: I completely missed updating this variable since it is farther down in the script. You can put this above the username and password check so its harder to miss. It would be worth it to add to the check as well, so users don't install the wrong EPEL version on their machines.

sthirugn:
@johnpmitsch updated

helpers/control_node_setup.sh
```diff
@@ -0,0 +1,40 @@
...
+export RHEL_MAJOR_VERSION="0"
+
+if [[ "$USERNAME" = "changeme" && "$PASSWORD" = "changeme" ]]; then
+  echo "Please update the variables in the script before running it."
+  exit 1
+fi
+if [ "$RHEL_MAJOR_VERSION" == "0" ]; then
+  echo "Please update the variables in the script before running it."
+  exit 1
+fi
```
johnpmitsch: @sthirugn can this be combined with the check on line 12?

sthirugn: I am mostly sure, we can.  But this part will go away in my next PR mostly :)

sthirugn:
I will create a new PR with more recent code.

sthirugn:
Since there were lot of changes in other parts of the code which made many part of this PR irrelevant, I chose to close this PR.  @johnpmitsch pointed out that it is not good practice to close the PR and open a new one - eventhough many parts of the PR are irrelevant and changed.

I updated this branch and tried to reopen the PR, but I am not able to do that here.  So I don't have an option other than submitting a new PR.  I will make sure to not close active PRs.

Issue #13, opened 2016-11-30
Fix: Satellite 6.1 installer is not picking correct interface


jyejare:
While installing Satellite 6.1, Its not picking the correct interface. The installer is throwing the error : 
```
Could not get the ip address from fact ipaddress_eth0 at /usr/share/katello-installer/modules/foreman_proxy/manifests/proxydhcp.pp:8 on node <hostname_here>
```
The bug is already logged for this issue : https://bugzilla.redhat.com/show_bug.cgi?id=1214627

To fix this issue, there is workaround mention below:
We have to explicitly pass the option --capsule-dhcp-interface <interface_name_on satellite>.

So can we fix this in our tasks/main.yml to use this option while installing 6.1 as:
```
- name: run Satellite 6.1 installer
  command: katello-installer --capsule-dhcp-interface {{ dhcp_interface }}        <------- Here !!
  when: satelliteversion == 6.1
```

sthirugn:
what is your satellite version?  The bug says it is fixed in Satellite 6.1 release.

jyejare:
My satellite version is 6.1.11 when installed with dhcp-interface option.

sghai:
This can be handled w/ ` ip -o link | grep -i 'state UP' | awk -F: '{print $2}'`

sghai:
Will raise a PR to address this

sthirugn:
@sghai are you planning to submit a pr for this?

sghai:
Yeah.. raised PR #63 w/ a change in cmd. Anove cmd won't work in case multiple interfaces are active. I tested that change w/ a good chunk of systems w/ all variations.

sthirugn:
Thank you Sachin.  But I like to have a more generic fix.  Because we won't know if the users (or customers) will use this with one interface or multiple interface.  What if we use your fix as a default option and let users override it with exposing new variables for capsule-dhcp-interface and capsule-dns-interface values in the vars/main.sample.yml file?

pgagne:
I believe ``--capsule-dhcp-interface`` sets the interface in ``/etc/sysconfig/dhcpd``. In my opinion we should give an option whether DHCP should configured or not, and if so with what interface.

sghai:
My PR covers just a workaround not a solution. Ideally this should be handled at product level. But I like the idea what @sthirugn suggested.

sthirugn:
[...]

PR #63, opened 2017-01-20
fetch default dhcp interface for 6.1

sghai:
handles #13

### DIFF
--- a/roles/sat6repro/tasks/main.yml
+++ b/roles/sat6repro/tasks/main.yml
@@ -78,8 +78,12 @@
 - name: untar config files
   command: tar --selinux --overwrite -xvf /tmp/config_files.tar.gz  -C /
   when: perform_restore
+- name: Identify the default dhcp interface
+  shell: route | grep -m 1 '^default' | awk '{print $8}'
+  register: dhcp_interface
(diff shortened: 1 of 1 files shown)


jyejare:
ACK, This should also remove DHCP_INTERFACE field from Satellite Upgrade job.

roles/sat6repro/tasks/main.yml
```diff
@@ -78,8 +78,12 @@
 - name: untar config files
   command: tar --selinux --overwrite -xvf /tmp/config_files.tar.gz  -C /
   when: perform_restore
+- name: Identify the default dhcp interface
+  shell: route | grep -m 1 '^default' | awk '{print $8}'
```
pgagne: I don't know if we want to assume that the default route interface is the one used for DHCP. This seems dangerous.

sthirugn:
https://github.com/RedHatSatellite/satellite-clone/issues/13#issuecomment-274080330

sthirugn:
@sghai please see https://github.com/RedHatSatellite/satellite-clone/issues/13#issuecomment-274955627

sghai:
thanks @sthirugn for updates.

PR #79, opened 2017-01-25
use unarchive module

johnpmitsch:


roles/sat6repro/tasks/restore.yml
```diff
@@ -13,9 +13,9 @@
 - name: Restore mongo data
   command: tar --selinux --overwrite -xvf {{ backup_dir }}/mongo_data.tar.gz -C /
 
-- name: Restore pulp data (this may take a long time!)
-  command: tar --selinux --overwrite -xvf {{ backup_dir }}/pulp_data.tar -C /
-  when: include_pulp_data
+- unarchive:
+    src: {{ backup_dir }}/pulp_data.tar
+    dest: /
```
sthirugn: can you test using `exclude` option to filter out: /var/lib/pulp/katello-export

sthirugn: see http://stackoverflow.com/a/38034804/3158707 if exclude doesn't work :)

sideangleside: Ansible complained loudly when I tested this PR with the following:

~~~
Always quote template expression brackets when they
start a value.
~~~

I ended up using the following
~~~
src: '{{ backup_dir }}/pulp_data.tar'
    dest: /
~~~

Lastly, since the backup needs to be local (to the system), I used `remote_src: True` as per the [Ansible Docs](https://docs.ansible.com/ansible/unarchive_module.html)

With that being said, I've run satellite-clone with this new option, and it is _significantly_ slower. When calling tar directly on my test system with a 334G pulp backup, it took about 60-80 minutes to restore. With the proposed, I am currently at 80 minutes with the CPU pegged at 100% running `usr/bin/gtar --diff -C / -f /data/Dump/satellite-backup/pulp_data.tar` and no pulp data has been restored yet.

johnpmitsch:
Closing this as it there are issues with the speed of the unarchive module

PR #82, opened 2017-01-25
Use ansible facts variable instead of `rhelversion` variable

sthirugn:


roles/sat6repro/tasks/main.yml
```diff
@@ -11,7 +11,7 @@
 - name: disable all repos
   command: subscription-manager repos --disable "*"
 - name: Enable required repos
-  command: subscription-manager repos --enable rhel-{{ rhelversion }}-server-rpms --enable rhel-server-rhscl-{{ rhelversion }}-rpms --enable rhel-{{ rhelversion }}-server-satellite-{{ satelliteversion }}-rpms
+  command: subscription-manager repos --enable rhel-{{ ansible_distribution_major_version }}-server-rpms --enable rhel-server-rhscl-{{ ansible_distribution_major_version }}-rpms --enable rhel-{{ ansible_distribution_major_version }}-server-satellite-{{ satelliteversion }}-rpms
```
johnpmitsch: This fact is coming from the destination node (and not from the control node)?

sthirugn: yes it comes from the destination node.  We are already using this variable in some places in main.yml

johnpmitsch: great :+1:

sthirugn:
Requesting review of this PR. This PR is open for so long.

PR #83, opened 2017-01-25
Added note saying rhel6 to rhel7 migration is not supported

sthirugn:


johnpmitsch:
I'm unsure if we have to mention anything about a migration (since that is a feature we came up with), but rather just mention that the destination node must be the same OS as the original satellite that the backup data came from.

sthirugn:
If not `migration`, what do you prefer to call it? - `porting`?

Or can you paste the exact sentence that you like to see?

docs/cloning.md
```diff
@@ -3,8 +3,9 @@
  **Important Notes:**
 
   - DHCP, DNS, and TFTP will be disabled during the install to avoid configuration errors. If you want to use provisioning on the cloned Satellite, you will have to manually re-enable these settings.
-  - The playbook will reset the admin password to "changeme"
+  - The playbook will reset the admin password to "changeme".
   - The playbook installs Satellite and may may take a while to complete.
+  - rhel6 to rhel7 migration of Satellite is not supported.  In other words, if your original Satellite is installed in rhel6, it can be cloned to a rhel6 vm only.
```
sthirugn: :+1:

sthirugn:
Updated as per the review comment.

PR #84, opened 2017-01-25
Created the root partition reallocation script

sthirugn:


docs/cloning.md
```diff
@@ -14,7 +14,7 @@
 
 #### On the Destination node (blank host)
 
-1. Make sure that the Destination node has adequate space and also make sure that the root partition has all the storage space. You may utilize this [script] (https://gist.githubusercontent.com/sthirugn/cdc34006ae280c344a15a474f7e35918/raw/d6382b8ffabcd9b2a5ab07150201388abfb7e01f/reallocate.sh) if needed.
+1. Make sure that the Destination node has adequate space and also make sure that the root partition has all the storage space. You may utilize this [script] (../helpers/reallocate.sh) if needed.
```
pgagne: @sthirugn perhaps this should be "You may utilize the script in ```<script-path``` instead of a link to the file?

sthirugn: do you mean like this?

https://raw.githubusercontent.com/satellite-clone/helpers/reallocate.sh

pgagne: My suggestion is to change this line to:

1. Make sure that the Destination node has adequate space and also make sure that the root partition has all the storage space. You may utilize the ```helpers/reallocate.sh``` script in the playbook for this.

pgagne:
ACK

PR #86, opened 2017-01-25
Make RH portal registration optional.

### LINKED ISSUE #73: Registration to the Customer Portal should be optional
As a user, I'd like for it to not be mandatory to register to the Customer Portal as a prerequisite of using satellite-clone. 

I may be using satellite-clone to test an upgrade or reproduce a problem, and I may want to do this offline or I may only have a single satellite subscription that I dont wish to remove from my production system. 

In this use case, I would expect that the user would be required to provide working yum repos (for RHEL, RHSCL & Satellite) prior to invoking the playbook.

sthirugn:
Fixes #73


sthirugn:
Requesting review of this PR. This is open for so long.

PR #96, opened 2017-01-31
Update README.md for control_node_setup script link

savitojs:
I was going through the documentation and found that link seems to be broken to setup control node. I've updated the link for the script. Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Ideally, you need two hosts to run this project:
    ```console
      # git clone https://github.com/RedHatSatellite/satellite-clone.git
    ```
-   NOTE: Optionally you may utilize the script [control_node_setup.sh] (../helpers/control_node_setup.sh) to perform step 2 below.  The instructions to use this script are documented in the script itself.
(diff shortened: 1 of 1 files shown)


sthirugn:
Hello @savitojs thank you for the PR.  The link here is not actually broken - for example - go to https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/cloning.md and click the hyperlink for `control_node_setup.sh` and it should work fine.

savitojs:
Hello @sthirugn Thank you. I tried to go to https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/cloning.md and didn't find `control_node_setup.sh` 

Actually I was reading the README.md and then clicked at that link, it says Page not found error 404.


Am I doing anything wrong ?

sthirugn:
Hello @savitojs I was totally looking in the wrong README file.  Sorry about that.  Although your suggested fix will solve the problem, it is not a good practice to point that doc to master always as it might change in other branches, tags, releases, etc.  The right fix for this issue is in https://github.com/RedHatSatellite/satellite-clone/pull/100.  Thank you again for the PR and for using this project.

savitojs:
Thank you @sthirugn @johnpmitsch

PR #98, opened 2017-01-31
Derive the hostname from the backup config files.

### LINKED ISSUE #21: Fetch the destination satellite hostname from the backup instead of asking it from the user
We should have a preliminary check that the hostname specified is the same as the one in the backups (for the clone playbook).

sthirugn:
- This makes the config even easier since the user doesn't need to fetch this manually.

Fixes #21

### DIFF
--- a/docs/cloning.md
+++ b/docs/cloning.md
@@ -2,6 +2,7 @@
 
  **Important Notes:**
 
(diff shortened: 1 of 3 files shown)


sthirugn:
I saw some backup files that do not have etc/foreman-proxy/settings.yml so I am putting this PR on hold for now until I find a more reliable solution.

sthirugn:
Rebased and ready for review/merge.

PR #99, opened 2017-01-31
Remove `perform_restore` option from configs.

### LINKED ISSUE #97: satellite-clone is intended only for cloning. Remove `perform_restore` option from the config.
background: The satellite-clone playbook was also used to quickly install a sat 6.1 or sat 6.2 box without actually restoring backup data.  But this beats the purpose of this playbook and this playbook should stay simple and support just cloning.  So `perform_restore` option will be removed.

sthirugn:
The satellite-clone playbook was also used to quickly install a sat 6.1 or sat 6.2 box without actually restoring backup data. But this beats the purpose of this playbook and this playbook should stay simple and support just cloning. So perform_restore option will be removed.  A new playbook needs to be created in future if there is an intent to support plain install of Satellite without backup data.

Fixes #97

### DIFF
--- a/roles/sat6repro/tasks/main.yml
+++ b/roles/sat6repro/tasks/main.yml
@@ -63,7 +63,6 @@
 # Run Satellite installer
 - name: untar config files
   command: tar --selinux --overwrite -xvf {{ backup_dir }}/config_files.tar.gz  -C /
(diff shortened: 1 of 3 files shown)


sthirugn:
Requesting review of this PR.

PR #106, opened 2017-02-01
Use a local yml file for user-defined variables

johnpmitsch:
This is a large PR so I tried to keep everything in separate commits to make it easier to review

I renamed the names of roles, playbooks, and docs to be consistent across the project.
This makes its easier for users and developers to update the files. The names of the playbooks should reflect the roles they are running especially since our playbooks run one role.

Users will edit the copied sample yml files in the top level of our
project and edit them as necessary. The vars files in the roles
of satellite-hostname and satellite-clone are used to set defaults
which are overrided by any other vars files or vars passed into
the playbook. This makes it so a user doesn't have to edit a role 
directly, and rather can override the defaults as they see fit.


johnpmitsch:
@ehelms if you would like to review the new file structure and use of local yml files, your input would be appreciated :)

roles/satellite-clone/tasks/main.yml
```diff
@@ -1,17 +1,22 @@
 ---
-- include_vars: main.yml
-
 # include pre_install_check
 - include: pre_install_check.yml
   when: run_pre_install_check
 
+- name: "check that variables are updated"
```
sthirugn: this can go inside pre_install_check.yml - all the other checks (including variables) are there

.gitignore
```diff
@@ -2,13 +2,11 @@
 *.retry
 *.tar.gz
 *.swp
-roles/sat6repro/files/*
-!roles/sat6repro/files/.gitkeep
+satellite-clone-vars.yml
+satellite-hostname-vars.yml
```
sthirugn: the above two can be moved under the section `#Varies per deployment`

roles/satellite-clone/tasks/main.yml
johnpmitsch: My thought behind putting it here instead of pre_install_check is that the pre_install_check can be skipped and this is a step that cannot be skipped (and is also easy to skip if a user just copies the file over w/o updating)

roles/satellite-clone/tasks/main.yml
```diff
@@ -1,10 +1,15 @@
...
+  fail: msg="Please update the variables in satellite-clone-vars.yml (after copying satellite-clone-vars.sample.yml)"
+  when:
+    - hostname == "satellite.example.com"
+    - activationkey == "changeme"
+    - org == "111111"
```
ehelms: I believe Ansible's when does a logical "and" here, which would imply that all of these conditions have to be true to fail. Thus, a user could change only one of these and get past this point. I think you want to swap to using "or" here?

roles/satellite-hostname/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+new_hostname: changeme
```
ehelms: Maybe change this to "example.com" ?

roles/satellite-hostname/tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: "check that variables are updated"
-  fail: msg="Please update the variables in roles/change_hostname/vars/main.yml"
+  fail: msg="Please update the variables in satellite-hostname-vars.yml"
```
ehelms: I may be wrong, but consider https://github.com/RedHatSatellite/satellite-clone/pull/106/files#r99473045 here as well.

roles/satellite-clone/tasks/main.yml
johnpmitsch: @ehelms my thought was that if they changed at least one, that confirms that they copied the local vars file over and edited it. But an "and" operator could be a more thorough check, provided a user wouldn't want to use a default (like satellite.example.com)

roles/satellite-hostname/tasks/main.yml
johnpmitsch: the satellite-hostname playbook is separate from the satellite-clone playbook, so they should just have to update `satellite-hostname-vars.yml`

.gitignore
```diff
@@ -2,13 +2,12 @@
...
+satellite-hostname.log
 
 #Varies per deployment
 inventory
-roles/sat6repro/vars/main.yml
-roles/change_hostname/vars/main.yml
+satellite-clone-vars.yml
+satellite-hostname-vars.yml
+
+
```
sthirugn: Empty lines can be removed

roles/satellite-clone/defaults/main.yml
```diff
@@ -0,0 +1,14 @@
...
+perform_restore: true
+backup_dir: /backup
+include_pulp_data: true
+activationkey: changeme
+org: 111111
+register_to_portal: true
+required_root_free_space: 75
+disable_firewall: false
+run_katello_reindex: false
+run_pre_install_check: true
```
sthirugn: For easy reference, the order of these variables should be the same as in satellite-clone-vars.sample.yml.

roles/satellite-clone/tasks/main.yml
sthirugn: In the case, can you update the task name to something like: `Mandatory check to verify if the required variables are updated` so it makes more sense to developers editing this in future.

roles/satellite-clone/tasks/main.yml
sthirugn: I vote for the `or` operator here which will benefit the user more although the usage of list (for and operator) is cool :)

roles/satellite-hostname/tasks/main.yml
sthirugn: +1 to @ehelms please use the `or` operator here too.

sthirugn:
in https://github.com/RedHatSatellite/satellite-clone/pull/106/files#diff-04b436162756b1b0bcfa9bf32e7d96e8L15: - s/satelliteversion/satellite_version

sthirugn:
Also rebase.

sthirugn:
test passed with one bug I noted above.

```sh
PLAY RECAP *********************************************************************
192.168.1.0                : ok=33   changed=18   unreachable=0    failed=0   

Restore pulp data (this may take a long time!) ----------------------- 1296.45s
run Satellite 6.2 installer ------------------------------------------ 1028.14s
Install Satellite 6.2 packages ---------------------------------------- 726.30s
Wait for foreman-tasks service to start ------------------------------- 300.02s
Restore mongo data ---------------------------------------------------- 114.29s
Stop katello services -------------------------------------------------- 59.94s
Restart katello-service ------------------------------------------------ 53.66s
Reset admin password --------------------------------------------------- 17.10s
Enable required repos -------------------------------------------------- 14.58s
register host ---------------------------------------------------------- 10.45s

Playbook finished: Mon Feb  6 19:47:11 2017, 51 total tasks.  1:01:25 elapsed. 
```

johnpmitsch:
@sthirugn @ehelms updated, I'll run another check to make sure this is still working

sthirugn:
https://github.com/RedHatSatellite/satellite-clone/pull/106/files#diff-04b436162756b1b0bcfa9bf32e7d96e8R28 is still pending to be updated.

johnpmitsch:
@sthirugn updated

sthirugn:
ACK. 

looks good, great work!

PR #113, opened 2017-02-08
Add FAQs

johnpmitsch:


docs/faqs.md
```diff
@@ -0,0 +1,23 @@
+# FAQs #
+
+##Cloning questions##
```
sthirugn: The word `questions` is redundant as we are already in FAQ page.  Instead, the name should match the role name for ease of use.

Maybe you can just name this as `satellite-clone`

docs/faqs.md
```diff
@@ -0,0 +1,23 @@
...
+
+2. Use the Satellite you are cloning as the Control node. You can clone the satellite-clone repo on the existing Satellite server and use that as the Control Node. The Destination Node is a blank machine.
+
+3. Use a separate machine as the Control Node. This is a machine that is seperate from the blank machine you are cloning to and the Satellite you are cloning. The backups will have to be copied onto the server. The Destination Node is a blank machine.
+
+####Can I clone Capsules?####
+
+Capsules are not able to be cloned at this time (Feb 2017). This is a planned feature.
+
+##Hostname change questions##
```
sthirugn: Title can be changed to `satellite-hostname`

docs/faqs.md
```diff
@@ -0,0 +1,23 @@
...
+
+1. Use the blank destination node as the Control node. This is the "one machine" approach. You can clone the satellite-clone repo on this server, move the backups to the specified folder on the server, and then run the playbook from the repo.
+
+2. Use the Satellite you are cloning as the Control node. You can clone the satellite-clone repo on the existing Satellite server and use that as the Control Node. The Destination Node is a blank machine.
+
+3. Use a separate machine as the Control Node. This is a machine that is seperate from the blank machine you are cloning to and the Satellite you are cloning. The backups will have to be copied onto the server. The Destination Node is a blank machine.
+
+####Can I clone Capsules?####
+
+Capsules are not able to be cloned at this time (Feb 2017). This is a planned feature.
```
sthirugn: What do you really mean when you say `capsule cloning`?  
- I think Capsule just syncs the content and does not store anything else isnt it?.  If so, is it not better to just resync the Capsule using the new cloned Satellite?

docs/faqs.md
johnpmitsch: The only concern with putting satellite-clone here is that is the name of the repo  so it may get confusing. We could do `satellite clone playbook` I suppose, but my thought around this FAQ is to put everything in layman's terms

docs/faqs.md
johnpmitsch: We are getting alot of questions about cloning a capsule, this is taking an existing capsule, and creating an duplicate one on another machine. Capsules do sync and store content on them to make the content available to clients.

docs/faqs.md
johnpmitsch: I still would like to keep this in ordinary, everyday english. I think what I could do is add the playbook names to `Cloning questions` and here, but still keep the titles. This may sort out any confusion

johnpmitsch:
@sthirugn updated

PR #117, opened 2017-02-15
Updated README with more info for Activation Key/Org in portal

sthirugn:


satellite-clone-vars.sample.yml
```diff
@@ -5,9 +5,18 @@
 satellite_version: 6.2
 
 # Registration info - Credentials to register your Satellite to Red Hat portal
-# You will need an activation key with a Satellite subscription
-# Activation keys can be created at https://access.redhat.com
-# For more information: https://access.redhat.com/management/activation_keys/new
+# - You will need an activation key with a Satellite subscription.
+# - For security reasons only Activationkey / Org combination is supported for now.
+#   (username/password combination is not supported).
```
johnpmitsch: I'm generally not in favor of pointing out a feature that hasn't been added. IMO it is saying to users "hey, you are missing out on this!" when they may not be thinking they are.

sthirugn:
Thank you @johnpmitsch I removed the line which mentioned about username/password being not supported

PR #132, opened 2017-03-20
Update docs for simplicity and uniformity

sthirugn:


docs/satellite-update.md
```diff
@@ -1,15 +1,16 @@
 ## Update Satellite to a new minor version
-#### **NOTE: THIS SHOULD ONLY BE USED FOR SATELLITE 6.2.X VERSIONS**
-1. Run the ansible playbook:
+### **Important warnings - read before running this playbook**
+ - This script must be used for Satellite 6.2 only.
```
johnpmitsch: If there is currently only one warning, do we really need a list and separate section?

sthirugn: I made a separate section to emphasize the fact this warning(s) should not be missed (similar to satellite-hostname.md).  But you are right, no need to have separate section though.

sthirugn:
@johnpmitsch ty, updated.

PR #144, opened 2017-03-24
Updated control node setup script

### LINKED ISSUE #135: Update control_node_setup.sh to remove fedora support

sthirugn:
Closed #135


johnpmitsch:
I'm not sure if we need a setup script anymore, there are really only two steps and we can put them in the README:
`yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm`
`yum install -y ansible`
It looks like the epel issue is being handled [here](https://git.io/vSkeW) and IIRC it only was an issue for RHEL 6, which we don't support anymore, so there is no need to remove it in a setup script

sthirugn:
@johnpmitsch I agree with your suggestion.  I will add the instruction to README and remove the helper script.

sthirugn:
Updated.

PR #150, opened 2017-03-27
Remove EPEL before satellite installer runs

### LINKED ISSUE #149: ensure EPEL is disabled
EPEL can conflict with the satellite installer, we should remove/disable it in the playbook

johnpmitsch:
Fixes #149

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -48,11 +48,14 @@
 - name: Ensure libselinux-python package is present (required by ansible)
   yum: name=libselinux-python state=latest
 
-# Remove epel-release package as it may cause error in satellite-installer run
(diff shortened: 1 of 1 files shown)


roles/satellite-clone/tasks/main.yml
```diff
@@ -48,11 +48,14 @@
...
-- name: remove epel-release package
-  yum: name=epel-release state=absent
-  when: ansible_distribution_major_version == "6"
+# Remove EPEL as it causes problems for the Satellite installer
+- name: Remove epel
+  yum_repository:
+    name: epel
+    state: absent
+
+- name: yum clean all
```
sthirugn: To be consistent with task names:
s/yum clean all/Clean yum info

sthirugn:
ty @johnpmitsch

PR #152, opened 2017-03-27
Fix mongo restore regression in migration scenario

### LINKED ISSUE #151: After running satellite-clone for rhel migration, pulp data is lost or corrupted

sthirugn:
Closed #151

### DIFF
--- a/roles/satellite-clone/tasks/restore.yml
+++ b/roles/satellite-clone/tasks/restore.yml
@@ -18,7 +18,7 @@
 
 - name: Restore mongo data
   command: "tar --selinux --overwrite -xf {{ backup_dir }}/mongo_data.tar.gz -C /"
-  when: standard_backup
+  when: rhel_migration or standard_backup
 
 - name: Start mongod
   service: name=mongod state=started


sthirugn:
@johnpmitsch rebased

roles/satellite-clone/tasks/restore.yml
```diff
@@ -18,7 +18,7 @@
 
 - name: Restore mongo data
   command: "tar --selinux --overwrite -xf {{ backup_dir }}/mongo_data.tar.gz -C /"
-  when: standard_backup
+  when: rhel_migration or standard_backup
```
johnpmitsch: I think we should change this to `when: mongo_data.stat.exists`

johnpmitsch: That would be similar to [katello-restore](https://git.io/vSYCS) logic

sthirugn: On second thought, I like this to be simple human readable format and like this to be directly correlating to https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/satellite-clone.md#prerequisites for easy maintenance in future.

So I like to keep it the same way `rhel_migration or standard_backup`

johnpmitsch: @sthirugn I think we should mirror katello-restore as much as possible, which works off of the files present rather than the scenario itself. We have checks (which admittedly, as we just discovered, need a little refining) to make sure one backup is present and then correct files are there

sthirugn: I am not sure why we should mirror katello-restore.  katello-restore doesn't handle migration or support finite defined use cases with proper error handling as satellite-clone. As long as satellite-clone makes cloning/migration easier, it is good.  I like to keep this tool as simple as possible to use and maintain.  In future, if there is a need to merge satellite-clone with katello-restore we can definitely think about it, but not right now.

Having said that:
Now pulp backup is the only file which is not dependent on the scenarios and we are rightfully handling it with:
```sh
- name: Restore pulp data
  command: "tar --selinux --overwrite -xf {{ backup_dir }}/pulp_data.tar --exclude=var/lib/pulp/katello-export -C /"
  when: pulp_data.stat.exists
```
All other tasks are clearly scenario dependent (see below).    It doesn't make sense to change mongo check until we change all scenarios listed below, which is not the intention of this PR.  If you feel strongly that this is the way to go, please create a new issue and we can brainstorm on how best to handle this.
```sh
- name: Restore postgresql data
  command: tar --selinux --overwrite -xf {{ backup_dir }}/pgsql_data.tar.gz -C /
  when: standard_backup

- include: restore_psql_dump.yml
  when: rhel_migration or online_backup

- name: Restore mongo data
  command: "tar --selinux --overwrite -xf {{ backup_dir }}/mongo_data.tar.gz -C /"
  when: rhel_migration or standard_backup

- include: restore_mongo_dump.yml
  when: online_backup
```

johnpmitsch:
@sthirugn 
a RHEL migration worked correctly with these changes
```
diff --git a/roles/satellite-clone/tasks/backup_check.yml b/roles/satellite-clone/tasks/backup_check.yml
index decf8e8..54c3c1c 100644
--- a/roles/satellite-clone/tasks/backup_check.yml
+++ b/roles/satellite-clone/tasks/backup_check.yml
@@ -62,16 +62,16 @@
     online_backup: True
   when: candlepin_dump.stat.exists and foreman_dump.stat.exists and mongo_dump.stat.exists
 
-- name: Check for standard backup
-  set_fact:
-    standard_backup: True
-  when: mongo_data.stat.exists and pgsql_data.stat.exists
-
 - name: Check for RHEL 6 to 7 migration backup
   set_fact:
     rhel_migration_backup: True
   when: mongo_data.stat.exists and foreman_dump.stat.exists and candlepin_dump.stat.exists
 
+- name: Check for standard backup
+  set_fact:
+    standard_backup: True
+  when: not rhel_migration_backup and mongo_data.stat.exists and pgsql_data.stat.exists
+
 - name: Check RHEL 6 to 7 migration backup is present
   fail:
     msg: |
```

roles/satellite-clone/tasks/restore.yml
johnpmitsch: @sthirugn ok lets keep it scenario dependent and fix that  "only one scenario exists" check in a separate PR

sthirugn: sounds good, does this mean ACK to this PR? :)

sthirugn:
ty @johnpmitsch

PR #158, opened 2017-03-28
Remove satellite_update role

### LINKED ISSUE #157: Remove satellite-update role from satellite-clone
Currently there are three roles:
- satellite-clone
- satellite-hostname
- satellite-update

The first two are related to cloning and migration. The third is a role to update satellite from one minor release to an other which does not make sense to maintain in this repository.  

Solution: Remove satellite-update from satellite-clone.  Add this to some other satellite maintenance project later as necessary.

sthirugn:
Fixes #157


sthirugn:
Thank you @johnpmitsch

PR #161, opened 2017-03-29
Fix permission issues and refactor postgresql_dump helper script

### LINKED ISSUE #159: Some issues with docs/helpers for RHEL6 to RHEL7
I was trying the RHEL6 to RHEL7 migration and ran into some hiccups:

The order of wording is confusing here:
https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/satellite-clone.md#prerequisites

Since it tells you to run katello-backup, then later tells you about  [satellite-clone/helpers/postgresql_dump.sh]((../helpers/postgresql_dump.sh)). That script seems to also run katello-backup. 

Does the foreman.dump stuff need to be created at the same time as the backup?

Also, the issues might run into some hiccups with that script. When run as root, the postgres user might not have permissions to write to the backup dir.

Do we either want to improve documentation here? or provide better tooling or both?

sthirugn:
Fixes #159

### DIFF
--- a/helpers/postgresql_dump.sh
+++ b/helpers/postgresql_dump.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Script to create backup files for rhel6 to 7 migration.
+# Script to create backup files for RHEL6 to 7 migration.
 #
 # Notes:
 # 1. This script runs katello-backup and also creates foreman.dump and candlepin.dump files from your Satellite6 server.
@@ -14,8 +14,6 @@ mkdir $backup_dir
 
 # Run katello-backup
 katello-backup $backup_dir
-echo "pgsql_data.tar.gz is not required for rhel6 to 7 migration..."
-rm -f $backup_dir/katello-backup*/pgsql_data.tar.gz
 
 # Create foreman and candlepin dumps
 katello-service stop
@@ -29,6 +27,8 @@ katello-service start
 
 set +x
 
-echo "Now copy the generated files to the destination host:"
-echo "config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar, "\
+echo "Now, copy the generated files to the destination host:"
+echo "config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar,"\
      "candlepin.dump, foreman.dump"
+echo "Note: Do not copy the generated pg_sql_data.tar.gz as this is not"\
+     "required for RHEL6 to 7 migration process."
--- a/helpers/postgresql_dump.sh
+++ b/helpers/postgresql_dump.sh
@@ -1,12 +1,34 @@
 #!/bin/bash
+# Script to create backup files for rhel6 to 7 migration.
+#
 # Notes:
-# 1. Use this script to create postgresql dumps - foreman.dump and candlepin.dump from the Satellite server.
-# 2. This scripts stops the katello-service in Satellite.  Please plan accordingly for the outage.
+# 1. This script runs katello-backup and also creates foreman.dump and candlepin.dump files from your Satellite6 server.
+# 2. Run this script as root user in Satellite6 server.
+# 3. This script stops the katello-service in Satellite.  Please plan accordingly for the outage.
+
+set -x
+
+# Create backup folder
 export backup_dir="/backup"
 mkdir $backup_dir
+
+# Run katello-backup
 katello-backup $backup_dir
+echo "pgsql_data.tar.gz is not required for rhel6 to 7 migration..."
+rm -f $backup_dir/katello-backup*/pgsql_data.tar.gz
+
+# Create foreman and candlepin dumps
 katello-service stop
 service postgresql start
-runuser - postgres -c "pg_dump -Fc foreman > $backup_dir/foreman.dump"
-runuser - postgres -c "pg_dump -Fc candlepin > $backup_dir/candlepin.dump"
+# Open up write access to let postgres user create dump files
+chmod o+w $backup_dir
+runuser - postgres -c "pg_dump -Fc foreman > "$backup_dir"/foreman.dump"
+runuser - postgres -c "pg_dump -Fc candlepin > "$backup_dir"/candlepin.dump"
+chmod o-w $backup_dir
 katello-service start
+
+set +x
+
+echo "Now copy the generated files to the destination host:"
+echo "config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar, "\
+     "candlepin.dump, foreman.dump"
--- a/docs/satellite-clone.md
+++ b/docs/satellite-clone.md
@@ -16,11 +16,7 @@
    Required backup files:
    - Standard backup scenario : config_files.tar.gz, mongo_data.tar.gz, pgsql_data.tar.gz, (optional) pulp_data.tar
    - Online backup scenario   : config_files.tar.gz, mongo_dump folder, foreman.dump, candlepin.dump, (optional) pulp_data.tar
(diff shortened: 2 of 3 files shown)


helpers/postgresql_dump.sh
```diff
@@ -1,12 +1,34 @@
...
+set -x
+
+# Create backup folder
 export backup_dir="/backup"
 mkdir $backup_dir
+
+# Run katello-backup
 katello-backup $backup_dir
+echo "pgsql_data.tar.gz is not required for rhel6 to 7 migration..."
+rm -f $backup_dir/katello-backup*/pgsql_data.tar.gz
```
johnpmitsch: The `katello-backup*` makes me nervous as the user may have multiple katello-backups and this will remove pgsql_data.tar.gz from all of them

sthirugn: your concern is valid, do you have an other suggestion?

johnpmitsch: @sthirugn the only thing I can think of is to grab the folder name from the second line of output here:
```
Starting backup: 2017-03-29 19:20:12 +0000
Creating backup folder /backup/katello-backup-2017-03-29T19:20:12+00:00
``` 
but then there is the issue of the special characters...

I'll think about it some more though

Maybe we can just print a reminder to the user to not copy the psql_data.tar.gz file  over when doing a RHEL migration

sthirugn: I am adding an explicit note to not copy the pgsql file.

johnpmitsch: @sthirugn if you want to include it, it looks like this will work
```
[root@sat-628-rhel7 /]# backup_dir=$(cat /tmp/backup_text.txt | awk '{w=$9} END{print w}')
[root@sat-628-rhel7 /]# cd $backup_dir
[root@sat-628-rhel7 katello-backup-2017-03-29T19:20:12+00:00]# 
```
That is just from the saved output from the file so you will have to replace that cat part and probably change the variable name since its  already used by you

johnpmitsch: @sthirugn this is working for me from the command line:
```
[root@sat-snap-rhel7 ~]# backup_subdir=$(katello-backup /backup | awk '{w=$9} END{print w}')
[root@sat-snap-rhel7 ~]# cd $backup_subdir
[root@sat-snap-rhel7 katello-backup-2017-03-29T19:41:43+00:00]#

Comment (sthirugn) [inline helpers/postgresql_dump.sh]: @johnpmitsch this works but relying on the word position of 9 in awk command may not be a reliable way to run in production.  I would rather leave it with a text warning.

Comment (sthirugn) [inline helpers/postgresql_dump.sh]: note that when katello-backup gets updated this position of the folder name can change and it is not safe.

File: helpers/postgresql_dump.sh [hunk 2]
```diff
@@ -29,6 +27,8 @@ katello-service start
 
 set +x
 
-echo "Now copy the generated files to the destination host:"
-echo "config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar, "\
+echo "Now, copy the generated files to the destination host:"
+echo "config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar,"\
      "candlepin.dump, foreman.dump"
+echo "Note: Do not copy the generated pg_sql_data.tar.gz as this is not"\
+     "required for migration process."
```
johnpmitsch: @sthirugn Please specify `RHEL 6 to 7 migration` here as `migration` is a bit ambiguous

helpers/postgresql_dump.sh
johnpmitsch: @sthirugn ya, I was thinking the same thing, I am ok with the warning, I am ok with the warning

johnpmitsch:
@sthirugn it looks fine other than my one comment - I'll try it out on a machine tomorrow

sthirugn:
@johnpmitsch thanks, updated your comment

helpers/postgresql_dump.sh
```diff
@@ -1,12 +1,34 @@
...
+chmod o-w $backup_dir
 katello-service start
+
+set +x
+
+echo "Now, copy the generated files to the destination host:"
+echo "config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar,"\
+     "candlepin.dump, foreman.dump"
+echo "Note: Do not copy the generated pg_sql_data.tar.gz as this is not"\
+     "required for RHEL6 to 7 migration process."
```
johnpmitsch: `pg_sql_data.tar.gz` should be `pgsql_data.tar.gz`

johnpmitsch:
@sthirugn The script ran correctly and without errors :+1:

After running the script, I am left with this file structure
```
/backup/
├── candlepin.dump
├── foreman.dump
└── katello-backup-2017-03-30T14:41:06+00:00
    ├── config_files.tar.gz
    ├── metadata
    ├── mongo_data.tar.gz
    ├── pgsql_data.tar.gz
    └── pulp_data.tar
```
The necessary files are split up into two different directories, what is expected of the user? Should they know that the candlepin.dump and foreman.dump files are associated with `katello-backup-2017-03-30T14:41:06+00:00`? I would also think that these files would be overwritten for each backup made.

This is making me reconsider the awk and find the filename from the output approach. The files could be moved to the correct backup folder, and the pgsql_data file can be removed. We can also display where the backup file is located at the end of the script output. Relying on the position of text output is not ideal, but I dont forsee that line changing in 6.2. We will also know about any upstream changes ahead of time and can handle those appropriately.

I'm open to alternative suggestions

helpers/postgresql_dump.sh
johnpmitsch: The user may not know where the backup files are located as the location displayed in the katello-backup output gets lost in the churn

johnpmitsch:
@sthirugn 
I think we can clean up this [prerequisites section](https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/satellite-clone.md#prerequisites) of the README. My thought is:
- rename this script to `rhel_6_to_7_migration_backup.sh` or something similar, since its explicitly used for generation backups for RHEL migrations
- Remove the additional notes section of that section, and make a "sub-bullet" under RHEL 6 to 7 migration scenario that tells the user to run this script to generate the appropriate backup
- The pulp data file being optional is made clear by the `(optional)` part in the list, we can either enhance that or leave it as is, but IMO no need to mention it again.

I bring this up as @pgagne mentioned he was a bit confused by that section, maybe he can comment with any clarifications as to what confused him

sthirugn:
@johnpmitsch thanks

- I addressed your comments mentioned in https://github.com/RedHatSatellite/satellite-clone/pull/161#issuecomment-290440629 - great suggestions btw
- For https://github.com/RedHatSatellite/satellite-clone/pull/161#issuecomment-290437934 -This PR is originally created for addressing permission issues in the script and it is already addressed.  katello-backup creating a new folder is a new issue introduced with 6.2.8 and needs to be addressed in another PR imho and should not block this PR.  I created a separate issue https://github.com/RedHatSatellite/satellite-clone/issues/164 to track this.

helpers/postgresql_dump.sh
sthirugn: valid concern.  User can use a new backup folder everytime.  This can be tracked and improved in https://github.com/RedHatSatellite/satellite-clone/issues/164

sthirugn:
@johnpmitsch as per the irc conv, the new output will look like:

```sh
Backup is created in /backup folder. Now copy the generated files to the destination host:
config_files.tar.gz, mongo_data.tar.gz, pulp_data.tar, candlepin.dump, foreman.dump
Note: Do not copy the generated pgsql_data.tar.gz as this is not required for RHEL6 to 7 migration process.
```

sthirugn:
Ty @johnpmitsch

PR #163, opened 2017-03-30
Add note that Capsules and Clients will have to reregister

### LINKED ISSUE #134: add ability to ensure cloned satellite is isolated from rest of environment
If you are using satellite-clone for testing upgrades, it is important that the cloned satellite is not able to talk to any other components of the production environment. The main thing to watch out for would be capsules but there might be other things as well.

Ideally, satellite-clone would have a flag to either block communication or remove any references to external services like capsules.

johnpmitsch:
Fixes #134
This is something we want to make sure that users are aware
of


sthirugn:
see https://github.com/RedHatSatellite/satellite-clone/issues/134#issuecomment-290463786

johnpmitsch:
closing in favor of https://github.com/RedHatSatellite/satellite-clone/pull/165

PR #165, opened 2017-03-30
Disassociate capsules with lifecycle environments

### LINKED ISSUE #134: add ability to ensure cloned satellite is isolated from rest of environment

johnpmitsch:
Fixes #134
The cloned satellite will still try to communicate with existing capsules,
so we can disassociate them to avoid any connection issues and interference
with exisiting setup. When the user has configured the capsules properly,
there will be instructions on how to reassociate the capsules.

### DIFF
--- a/docs/satellite-clone.md
+++ b/docs/satellite-clone.md
@@ -9,6 +9,7 @@
   - You can clone RHEL 6 backup data to a RHEL 7 machine.  In this case, you must update the variable `rhel_migration` to true as explained later in this document. Please note that this scenario is supported only for *Satellite 6.2*.
   - If you are using NFS for storage and your pulp backup tar file is large (>150 gb), you might see memory errors while untaring pulp data.  In this case you can optionally choose to skip pulp restore (by setting `include_pulp_data` to `false` in `satellite-clone-vars.yml`).
   - After running the playbook, existing manifests may have to be refreshed.
+  - Capsules will be unassociated with Lifecycle environments to avoid any interference with existing infrastructure. Instructions to reverse these changes can be found in `logs/reassociate_capsules.txt` under Satellite Clone's root directory
 
 #### Prerequisites ####
 
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -1,0 +1,27 @@
+#!/usr/bin/ruby
+username = "admin"
+password = "changeme"
+external_capsules = []
+external_capsule_ids = `hammer -u #{username} -p #{password} --csv capsule list --search 'feature = \"Pulp Node\"' | tail -n+2 | awk -F, {'print $1'}`
+external_capsule_ids.split("\n").each do |id|
+  lifecycle_environment = `hammer -u #{username} -p #{password} --csv capsule content lifecycle-environments --id #{id} | tail -n+2 | awk -F, {'print $2'}`.split("\n")
+  name = `hammer -u #{username} -p #{password} --csv capsule info --id #{id} | tail -n+2 | awk -F, {'print $2'}`.chomp
+  organization = `hammer -u #{username} -p #{password} --csv capsule info --id 2 | tail -n+2 | awk -F, '{print $(NF-2)}'`.chomp
+  external_capsules << {:id => id, :name => name, :lifecycle_environments => lifecycle_environment, :organization => organization}
+end
+
+reverse_commands = []
+external_capsules.each do |capsule|
+  capsule[:lifecycle_environments].each do |env|
+    `hammer -u #{username} -p #{password} --csv capsule content remove-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization "#{capsule[:organization]}"`
+    reverse_command = "hammer -u #{username} -p #{password} --csv capsule content add-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization \"#{capsule[:organization]}\""
+    reverse_commands << reverse_command
+  end
+end
+
+STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
+            "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
+            " making sure to replace the credentials with your own."
+reverse_commands.each do |reverse|
+  STDOUT.puts reverse
+end
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -147,6 +147,13 @@
   command: satellite-installer --upgrade
   when: rhel_migration
 
+- name: Disassociate capsules with lifecycle environments (to avoid the cloned Satellite server talking with live capsules)
+  script: disassociate_capsules.rb
+  register: disassociate_capsules
+
+- name: copy disassociate_capsules.rb output
+  local_action: copy content={{ disassociate_capsules.stdout }} dest={{ playbook_dir }}/logs/reassociate_capsules.txt
+
 - debug:
     msg: "****NOTE**** Your Satellite's hostname is updated to match the original Satellite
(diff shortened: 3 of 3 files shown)


johnpmitsch:
I tested this out on a backup with two capsules associated with multiple lifecycle envs, The capsules were no longer associated with an environment on a clone, and the logs displayed these instructions for reassociating them:
```
All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite and any interference with existing infrastructure. To reverse these changes, run the following commands:
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 3 --environment Library --organization "Default Organization"
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 3 --environment dev --organization "Default Organization"
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 3 --environment prod --organization "Default Organization"
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 2 --environment Library --organization "Default Organization"
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 2 --environment dev --organization "Default Organization"
```

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/ruby
+username = "admin"
+password = "changeme"
```
bbuckingham: Does username/password need to be configurable?  Or can we assume that it will always be these values?

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -0,0 +1,26 @@
...
+external_capsules = []
+external_capsule_ids = `hammer -u #{username} -p #{password} --csv capsule list --search 'feature = \"Pulp Node\"' | tail -n+2 | awk -F, {'print $1'}`
+external_capsule_ids.split("\n").each do |id|
+  lifecycle_environment = `hammer -u #{username} -p #{password} --csv capsule content lifecycle-environments --id #{id} | tail -n+2 | awk -F, {'print $2'}`.split("\n")
+  name = `hammer -u #{username} -p #{password} --csv capsule info --id #{id} | tail -n+2 | awk -F, {'print $2'}`.chomp
+  organization = `hammer -u admin -p changeme --csv capsule info --id 2 | tail -n+2 | awk -F, '{print $(NF-2)}'`.chomp
```
bbuckingham: May want to update to use the username/password variables.

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -0,0 +1,26 @@
...
+  external_capsules << {:id => id, :name => name, :lifecycle_environments => lifecycle_environment, :organization => organization}
+end
+
+reverse_commands = []
+external_capsules.each do |capsule|
+  capsule[:lifecycle_environments].each do |env|
+    `hammer -u admin -p changeme --csv capsule content remove-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization "#{capsule[:organization]}"`
```
bbuckingham: Same here (use of variables).

roles/satellite-clone/files/disassociate_capsules.rb
johnpmitsch: @bbuckingham We reset the password to admin/changeme [here](https://git.io/vS4xlZ) so it will be these values at this point in the playbook

roles/satellite-clone/files/disassociate_capsules.rb
johnpmitsch: good catch, will update

roles/satellite-clone/files/disassociate_capsules.rb
johnpmitsch: Good catch again :)

johnpmitsch:
@bbuckingham @sthirugn  A test run on Dolly completed successfully (5 passing scenarios + 1 expected failure w/ incorrect backup).
I updated the script according to suggestions, I'll test the new script to make sure its ok

johnpmitsch:
script is working ok on machine w/ capsules

docs/satellite-clone.md
```diff
@@ -9,6 +9,7 @@
   - You can clone RHEL 6 backup data to a RHEL 7 machine.  In this case, you must update the variable `rhel_migration` to true as explained later in this document. Please note that this scenario is supported only for *Satellite 6.2*.
   - If you are using NFS for storage and your pulp backup tar file is large (>150 gb), you might see memory errors while untaring pulp data.  In this case you can optionally choose to skip pulp restore (by setting `include_pulp_data` to `false` in `satellite-clone-vars.yml`).
   - After running the playbook, existing manifests may have to be refreshed.
+  - Capsules will be unassociate with Lifecycle environments to avoid any interference with existing infrastructure. Instructions to reverse these changes can be found in `logs/reassociate_capsules.txt` under Satellite Clone's root directory
```
sthirugn: s/unassociate/unassociated

roles/satellite-clone/files/disassociate_capsules.rb
sthirugn: It is better to add a comment saying that the clone playbook will change the credentials to admin/changeme.  If the user had changed the password after cloning, before running this script - then these credentials need to be updated accordingly.

johnpmitsch: @sthirugn I'm not sure if I understand your concern - This script is meant to be used in the playbook only and its not exposed to the user in any way, so its a safe assumption that the credentials will be admin/changeme when the script is run. We also mention the credential change in the docs [here](https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/satellite-clone.md#important-notes)

johnpmitsch:
@sthirugn updated typo

roles/satellite-clone/files/disassociate_capsules.rb
sthirugn: It may not be a safe assumption.  Here's why:
My user scenario is:
- clone the satellite (didnt pay lot of attention to the terminal instructions about capsule)
- change satellite's admin password as this is one of the very first things which admins (are supposed to) do
- Realize that capsules are disassociated, reviewing the logs, finding the command to reset capsule, run it - boom!

sthirugn: Never mind, I thought this is a script to reassociate. sorry about that.

johnpmitsch: @sthirugn ok, I understand, you are saying that the commands that are outputted in the logs will have the same username/password (even after they changed it) and there should be a note in the logs, I can change that.

roles/satellite-clone/tasks/main.yml
```diff
@@ -147,6 +147,13 @@
   command: satellite-installer --upgrade
   when: rhel_migration
 
+- name: Disassociate capsules with lifecycle environments
```
sthirugn: May be it is better to add more info in the task name, something like:

`name: Disassociate capsules with lifecycle environments (to avoid cloned Satellite server talking to live Capsules)`

sthirugn:
I just have one other minor comment.

roles/satellite-clone/files/disassociate_capsules.rb
sthirugn: yeah that will be great

johnpmitsch: @sthirugn It creates the reversal commands as well, I updated with a line about changing your credentials before running

johnpmitsch:
@sthirugn updated

sthirugn:
@johnpmitsch thanks!

PR #173, opened 2017-04-07
don't check password is changeme on hostname change

johnpmitsch:
Our default password is "changeme" in dev environments and after a clone, so this check fails quite a bit. Just checking the new_hostname and admin_username have changes is enough to tell if the user has edited the file.

We should probably refine this check a bit more, but this will unblock automation for now


sthirugn:
ACK

PR #188, opened 2017-04-25
adds wrapper script

johnpmitsch:
This is a wrapper script that will allow us to
use a cli tool when this repo is packaged. It
works off of the yml config files that ansible
uses in order to not duplicate the parameters
and documentation we have in those files

### DIFF
--- a/packaging/satellite-clone.rb
+++ b/packaging/satellite-clone.rb
@@ -1,0 +1,71 @@
+#!/usr/bin/env ruby
+require 'optparse'
+
+DEFAULT_PRODUCTION_INSTALL_PATH = "/usr/share/satellite-clone"
+DEFAULT_PLAYBOOK_FILE = DEFAULT_PRODUCTION_INSTALL_PATH + "/" + "satellite-clone-playbook.yml"
+
+@options = {}
+@options[:development] = false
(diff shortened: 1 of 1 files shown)


johnpmitsch:
To replicate what the spec file will do - first checkout the PR, then:
```
mkdir /etc/satellite-clone
cp satellite-clone-vars.sample.yml /etc/satellite-clone/satellite-clone-vars.yml
ln -s /etc/satellite-clone/satellite-clone-vars.yml satellite-clone-vars.yml
yum install -y ruby ansible
```
To do a full clone, add backup files to /backup and update /etc/satellite-clone/satellite-clone-vars.yml
Then to run the playbook:
```
cd /usr/share/satellite-clone
./packaging/satellite-clone.rb
```
On a yum install, the user will just run `satellite-clone`
There also is a development mode, run by `./packaging/satellite-clone --development` which will run satellite-clone in the current directory, instead of running from /usr/share

sthirugn:
I am not much familiar with ruby.  But technically it looks correct.  I will let Brad or someone review this.

bbuckingham:
The logic looks good.  I haven't gotten to run through testing; however, I do not want to hold up making this available for more usage.  ACK!

Thanks @johnpmitsch !

PR #197, opened 2017-05-08
Mention that a subscription is required

johnpmitsch:


sthirugn:
ACK

bbuckingham:
ACK

sthirugn:
actually lets add a note in both the places
https://github.com/RedHatSatellite/satellite-clone/blob/master/docs/satellite-clone.md#cloning-a-satellite-server
https://github.com/RedHatSatellite/satellite-clone#what-you-need

PR #201, opened 2017-05-24
Use mongo_dump folder for RHEL migration

### LINKED ISSUE #202: restoring mongo_data.tar.gz does not work on RHEL migration
Error is `mongod child process failed, exited with error number 100` on starting mongo

This could be because of the difference in mongo versions.

We can use mongo_dump folder to restore instead of mongo_data.tar.gz

johnpmitsch:
Fixes #202 
- [ ] refactor to eliminate repetitive code since the online backup and RHEL migration are the same


sthirugn:
I am trying to understand why it worked earlier and not now. Can you provide more info on the following?

- Does this affect standard cloning scenario? If so, we should start using mongo dump for all scenarios.
- Where do you actually see the difference in mongo versions? We didnt ship any new mongo version in recent 6.1 or 6.2 z-streams afaik

johnpmitsch:
@sthirugn 
standard cloning scenario is from a standard backup, so it would have to use mongo_data.tar.gz, mongo_dump is only from an online backup. In this case, the issue affected a RHEL 6 to 7 migration. Talking with @dkliban from pulp, he informed me that mongo_dump is the more reliable solution since mongo will handle restoring the data opposed to untaring mongo files to the filesystem. Since we are requiring foreman.dump and candlepin.dump from the online backup for rhel migration, it makes sense to use the mongo_dump folder that is created as well,  since that is more reliable of the two options and requires less work to generate.

The mongo version was different between:
RHEL6 satellite-6.2.0-21.1.el6sat.noarch - mongodb-server-2.4.6-2.el6sat.x86_64
RHEL7 satellite-6.2.9-7.0.el7sat.noarch - mongodb-2.6.11-2.el7sat.x86_64

from talking with Zach on #satellite6, it sounds like the difference of mongo versions is in sat 6.2.z (possibly upgraded in a very early z version)

johnpmitsch:
@sthirugn correction - the mongo versions are actually different between RHEL 6 + RHEL 7, but no difference in 6.2.z versions.
from #satellite6
```
jomitsch | zhunting crag: ah ok, so to confirm, RHEL 6 - 2.4.6-2, RHEL 7 - 2.6.11-2?
zhunting | yeah                                                                     
```

sthirugn:
Thank you @johnpmitsch. Based on your response, mongo did not change in rhel6 or rhel7 satellite z-stream for a while now.

So I am still not able to understand - Why our tests for migration of rhel6 to 7 passed between Feb till May but not now.  Is this setup having anything different from the test setups we had earlier? I am trying to understand the root cause so we can provide a permanent reliable fix.

johnpmitsch:
@sthirugn I don't think this is a "fail every case" type of issue, and I also can't say for sure that it's related to the mongo versions. But if we have a more reliable solution of restoring mongo data, that fixed a bug in this real-world scenario, I think the question more is why not use it?

sthirugn:
I discussed with @johnpmitsch in call. This scenario was tested many times earlier but we dint see this error (may be because of the available content types).  I agree that taking care of mongo version is good for the overall reliability of the tool.

roles/satellite-clone/tasks/backup_check.yml
```diff
@@ -97,4 +91,4 @@
       Standard backup scenario - config_files.tar.gz, mongo_data.tar.gz, pgsql_data.tar.gz, (optional) pulp_data.tar
       Online backup scenario - config_files.tar.gz, mongo_dump folder, foreman.dump, candlepin.dump, (optional) pulp_data.tar
```
sthirugn: this row and the next needs to be merged now as online backup and rhel6 to 7 migration backup are using similar backup files.

roles/satellite-clone/tasks/backup_check.yml
```diff
@@ -97,4 +91,4 @@
...
       RHEL 6 to 7 migration scenario - config_files.tar.gz, mongo_data.tar.gz, foreman.dump, candlepin.dump, (optional) pulp_data.tar
```
sthirugn: mongo_data.tar.gz should be changed to folder here and this line can be merged with the previous one.

sthirugn:
Lines 79 and 80 need to be fixed as well.
```
      Backup is missing files. For an online backup, you need candlepin, foreman, and
      mongo dump files. For an online backup, you need candlepin.dump, foreman.dump,
      and mongo_dump folder. For a standard (offline) backup, you will need mongo_data.tar.gz
```

sthirugn:
where is the actual restoring of mongo_dump folder happens in restore process of rhel migration scenario?

johnpmitsch:
@sthirugn [here](https://git.io/vQaS8)

sthirugn:
Thanks it makes sense.

sthirugn:
ACK great work!

PR #205, opened 2017-07-05
Remove hostname change playbook

johnpmitsch:
Sat-clone shouldn't handle the hostname change as the script is being
included as an official tool in Satellite 6.2.12+


johnpmitsch:
The current upstream script (which the playbook uses) is also not compatible with satellite 6.2

sthirugn:
ACK

PR #216, opened 2017-09-26
add option to run pg_restore

### LINKED ISSUE #215: add option to run pg_restore with --disable-triggers
If there is existing data that breaks sql constraints in the schema, the pg_restore command will error out with something like: 
```
stderr: pg_restore: [archiver (db)] Error while PROCESSING TOC:
pg_restore: [archiver (db)] Error from TOC entry 4718; 2606 22417 CONSTRAINT remote_execution_features_pkey foreman
pg_restore: [archiver (db)] could not execute query: ERROR:  could not create unique index "remote_execution_features_pkey"
DETAIL:  Key (id)=(7) is duplicated.
    Command was: ALTER TABLE ONLY remote_execution_features
    ADD CONSTRAINT remote_execution_features_pkey PRIMARY KEY (id);
```

This can be avoided using the --disable-triggers option. We should make an option to enable this in case users run into this.

johnpmitsch:
Fixes #215

### DIFF
--- a/roles/satellite-clone/defaults/main.yml
+++ b/roles/satellite-clone/defaults/main.yml
@@ -9,3 +9,4 @@ run_katello_reindex: False
 run_pre_install_check: True
 register_to_portal: True
 rhel_migration: False
+disable_postgres_triggers: False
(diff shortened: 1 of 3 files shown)


johnpmitsch:
Tested on 6.2 rhel7 backup, worked well.

PR #235, opened 2017-10-18
Disable REX on install

### LINKED ISSUE #234: Recommend cloning to isolated env and disable REX on clone
A production satellite can communicate to clients through manual errata updates and automated REX plans. We should do two things to ensure the new cloned satellite doesn't interfere with a user's existing production satellite clients:

- Recommend that users clone to an isolated network if they are looking to use their clone for non-production "sandbox" purposes. This is defined as a network that can't talk to their original Satellite's network.
-  Disable REX plugin.

johnpmitsch:
Fixes #234 

To do:
- [x] update docs to recommend cloning to isolated env

### DIFF
--- a/docs/satellite-clone.md
+++ b/docs/satellite-clone.md
@@ -3,13 +3,14 @@
 #### Important Notes ####
 
   - The playbook will update the cloned Satellite hostname to match the hostname of the original Satellite from which the backup is generated.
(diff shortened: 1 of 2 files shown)


johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

PR #245, opened 2017-10-24
Updates production cli

### LINKED ISSUE #229: expose "resume at step" and/or "interactive step" functionality for sat-clone
also worth noting we need a man page

http://docs.ansible.com/ansible/latest/playbooks_startnstep.html

### LINKED ISSUE #241: Hide deprecation warnings in sat-clone production
we can somehow add deprecation_warnings=False to ansible.cfg

johnpmitsch:
Fixes #241
Hide deprecation warnings in prod

Fixes #229
expose "resume at step" and "interactive step" functionality

also removes development mode as its unnecessary

Updated variables to use cacheable facts so we can start at any task after a full run and the
variables will be saved. I added some developer notes about this in the docs as well.

The cli executable has been updated to use python instead of ruby to reduce a dependency
(ansible requires python already)
TODO
- [x] add development notes to README around using facts for caching
- [x] test actual rpm build
- [x] change spec script in packaging


johnpmitsch:
Resume at step and interactive mode actually isn't going to work for us currently as we rely on alot of facts and registered variables in the playbook run.

We may be able to do something with fact caching http://docs.ansible.com/ansible/latest/playbooks_variables.html#fact-caching

I think any `register` would have to change to `set_fact` in this case.

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

sthirugn:
looks good to me but I don't have time to test the new changes.

johnpmitsch:
An example of `--start-at-task` functionality below. This was run after a successful playbook run so the facts were cached:
```
[root@sat-clone-test2 ~]# satellite-clone --start-at-task="run Satellite 6.2 installer"
This will initiate satellite-clone. Do you want to proceed? (y/n): y
ansible-playbook -i /usr/share/satellite-clone/inventory --start-at-task "run Satellite 6.2 installer" /usr/share/satellite-clone/satellite-clone-playbook.yml
[WARNING]: log file at /var/log/satellite-clone/playbook.log is not writeable and we cannot create it, aborting



PLAY [satellite6] **********************************************************************************************************************************************************



TASK [satellite-clone : run Satellite 6.2 installer] ***********************************************************************************************************************

changed: [localhost]



cmd: 

satellite-installer

--scenario

satellite

--foreman-proxy-dns

false

--foreman-proxy-dhcp

false

--foreman-proxy-tftp

false

--no-enable-foreman-plugin-remote-execution

--no-enable-foreman-proxy-plugin-remote-execution-ssh



start: 2017-11-06 19:17:02.950438



end: 2017-11-06 19:18:38.348100



delta: 0:01:35.397662



stdout:   Success!

... (playbook continues)


```

And /usr/share/satellite-clone/cache/localhost contains (among other facts)
```
    "hostname": "sat-snap-rhel7.example.com", 
    "module_setup": true, 
    "online_backup": true, 
    "root_free_space": "95.377412096", 
    "satellite_version": 6.2, 
    "standard_backup": false
```

johnpmitsch:
yee-haw

johnpmitsch:
@ehelms can you take a glance at the satellite-clone.py file and the ansible configs since you helped me come up with this implementation?

ansible.production.cfg
```diff
@@ -0,0 +1,10 @@
+[defaults]
+callback_plugins = /usr/share/satellite-clone/callback_plugins/
+callback_whitelist = profile_tasks
+inventory = /usr/share/satellite-clone/inventory
+log_path = /var/log/satellite-clone/playbook.log
+gathering = smart
+fact_caching = jsonfile
+fact_caching_connection = /usr/share/satellite-clone/cache
+fact_caching_timeout = 259200
+deprecation_warnings=False
```
ehelms: Nitpick of putting spaces around `=`

ehelms:
LGTM but I've asked @evgeni to check the python CLI part as well

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+            return True
+        if reply[0] == 'n':
+            exit()
+
+def tail(f, n, offset=0):
+  offset_total = str(n+offset)
+  stdin,stdout = os.popen2("tail -n "+offset_total+" "+f)
+  stdin.close()
+  lines = stdout.readlines(); stdout.close()
+  return lines[:,-offset]
```
ehelms: Indentation off here

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+
+    inventory_path = "{0}/inventory".format(DEFAULT_PRODUCTION_INSTALL_PATH)
+    ansible_playbook = "ansible-playbook -i {0} {1} {2}".format(inventory_path, 
+                                                                 " ".join(ansible_args),
+                                                                 DEFAULT_PLAYBOOK_FILE)
+ 
+    print(ansible_playbook)
+    run_playbook = subprocess.Popen(ansible_playbook, 
+                                    shell=True, 
+                                    stdout=subprocess.PIPE,
```
ekohl: By default the process' stdout is piped to stdout so I don't think the real-time output code below is neded.

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+    ansible_playbook = "ansible-playbook -i {0} {1} {2}".format(inventory_path, 
+                                                                 " ".join(ansible_args),
+                                                                 DEFAULT_PLAYBOOK_FILE)
+ 
+    print(ansible_playbook)
+    run_playbook = subprocess.Popen(ansible_playbook, 
+                                    shell=True, 
+                                    stdout=subprocess.PIPE, 
+                                    bufsize=1,
+                                    cwd="{0}".format(DEFAULT_PRODUCTION_INSTALL_PATH))
```
ekohl: Is the string formatting really needed? Can't you just use ```cwd=DEFAULT_PRODUCTION_INSTALL_PATH```?

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+        ansible_args.append("--interactive")
+
+    if not os.path.isdir(DEFAULT_PRODUCTION_INSTALL_PATH):
+         print("It looks like satellite-clone has not been installed properly, " \
+               "{0} does not exist.".format(DEFAULT_PRODUCTION_INSTALL_PATH))
+         exit()
+
+    os.environ["ANSIBLE_CONFIG"] = "{0}/ansible.production.cfg".format(DEFAULT_PRODUCTION_INSTALL_PATH)
+
+    inventory_path = "{0}/inventory".format(DEFAULT_PRODUCTION_INSTALL_PATH)
```
ekohl: ```os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, 'inventory')``` is cleaner.

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+        ansible_args.append("--start-at-task \"{0}\"".format(options.start_at_task))
+    if options.interactive:
+        ansible_args.append("--interactive")
+
+    if not os.path.isdir(DEFAULT_PRODUCTION_INSTALL_PATH):
+         print("It looks like satellite-clone has not been installed properly, " \
+               "{0} does not exist.".format(DEFAULT_PRODUCTION_INSTALL_PATH))
+         exit()
+
+    os.environ["ANSIBLE_CONFIG"] = "{0}/ansible.production.cfg".format(DEFAULT_PRODUCTION_INSTALL_PATH)
```
ekohl: ```os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, 'ansible.production.cfg')``` is cleaner.

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+    while "the answer is invalid":
+        reply = str(raw_input(question+' (y/n): ')).lower().strip()
+        if reply[0] == 'y':
+            return True
+        if reply[0] == 'n':
+            exit()
+
+def tail(f, n, offset=0):
+  offset_total = str(n+offset)
+  stdin,stdout = os.popen2("tail -n "+offset_total+" "+f)
```
ekohl: Would it equal and simpler to use:
```python
lines = subprocess.check_call(["tail", "-n", n + offset, f])
return lines[:,-offset]
```

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+
+from optparse import OptionParser
+import os
+import subprocess
+
+def yes_or_no(question):
+    while "the answer is invalid":
```
evgeni: You mean `while True`? :)

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+
+from optparse import OptionParser
```
evgeni: `optparse` is [deprecated in 2.7](https://docs.python.org/2/library/optparse.html), and one should use [`argparse`](https://docs.python.org/2/library/argparse.html) instead. AFAIK `satellite-clone` is only supported on EL7, so depending on Python 2.7 should be fine?

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+
+def yes_or_no(question):
+    while "the answer is invalid":
+        reply = str(raw_input(question+' (y/n): ')).lower().strip()
+        if reply[0] == 'y':
+            return True
+        if reply[0] == 'n':
+            exit()
+
+def tail(f, n, offset=0):
```
ekohl: This function is unused?

ehelms:
A few thoughts you might consider:

 * add pylint or pytestcode via travis
 * add ansible-lint via travis

bin/satellite-clone.py
evgeni: What's `f` and `n`?

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+  offset_total = str(n+offset)
```
evgeni: Python uses 4 space indentation.

bin/satellite-clone.py
evgeni: [`subprocess.check_output`](https://docs.python.org/2/library/subprocess.html#subprocess.check_output)

bin/satellite-clone.py
evgeni: Uh, what is this line supposed to do?

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if not os.path.isdir(DEFAULT_PRODUCTION_INSTALL_PATH):
+         print("It looks like satellite-clone has not been installed properly, " \
+               "{0} does not exist.".format(DEFAULT_PRODUCTION_INSTALL_PATH))
+         exit()
+
+    os.environ["ANSIBLE_CONFIG"] = "{0}/ansible.production.cfg".format(DEFAULT_PRODUCTION_INSTALL_PATH)
+
+    inventory_path = "{0}/inventory".format(DEFAULT_PRODUCTION_INSTALL_PATH)
+    ansible_playbook = "ansible-playbook -i {0} {1} {2}".format(inventory_path,
```
evgeni: `subprocess.Popen` supports `args` being a list, which is so much nicer than playing with `.format` and `.join` :)

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+    os.environ["ANSIBLE_CONFIG"] = "{0}/ansible.production.cfg".format(DEFAULT_PRODUCTION_INSTALL_PATH)
+
+    inventory_path = "{0}/inventory".format(DEFAULT_PRODUCTION_INSTALL_PATH)
+    ansible_playbook = "ansible-playbook -i {0} {1} {2}".format(inventory_path, 
+                                                                 " ".join(ansible_args),
+                                                                 DEFAULT_PLAYBOOK_FILE)
+ 
+    print(ansible_playbook)
+    run_playbook = subprocess.Popen(ansible_playbook, 
+                                    shell=True,
```
evgeni: I don't think you need `shell=True` here, no?

bin/satellite-clone.py
evgeni: `os.path.join`

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+    parser.add_option("-i", "--interactive",
+                      action="store_true",
+                      dest="interactive",
+                      default=False,
+                      help="interactive mode with prompt for each task")
+    (options, args) = parser.parse_args()
+
+
+    DEFAULT_PRODUCTION_INSTALL_PATH = "/usr/share/satellite-clone"
+    DEFAULT_PLAYBOOK_FILE = DEFAULT_PRODUCTION_INSTALL_PATH + "/satellite-clone-playbook.yml"
```
evgeni: `os.path.join`

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+    ansible_args = [] 
+    if options.verbose:
+        ansible_args.append("-vvv")
+    if options.start_at_task:
+        ansible_args.append("--start-at-task \"{0}\"".format(options.start_at_task))
+    if options.interactive:
+        ansible_args.append("--interactive")
+
+    if not os.path.isdir(DEFAULT_PRODUCTION_INSTALL_PATH):
+         print("It looks like satellite-clone has not been installed properly, " \
```
evgeni: where is your `from __future__ import print_function`? otherwise it's `print "string"` ;)

bin/satellite-clone.py
```diff
@@ -0,0 +1,83 @@
...
+               "{0} does not exist.".format(DEFAULT_PRODUCTION_INSTALL_PATH))
+         exit()
```
evgeni: You want to exit 1 here, or similar.

bin/satellite-clone.py
johnpmitsch: Thanks, I'll test it out w/o the code block below

johnpmitsch:
@ekohl @ehelms thanks for the review!

bin/satellite-clone.py
ekohl: technically a non-empty string is trueish so it will work :)

bin/satellite-clone.py
johnpmitsch: I'm pretty sure I needed to add that for this part to function properly

bin/satellite-clone.py
johnpmitsch: having the string does explicitly explain the condition of the loop, maybe keep it for this reason?

bin/satellite-clone.py
```diff
@@ -0,0 +1,76 @@
...
+
+    os.environ["ANSIBLE_CONFIG"] = os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, 'ansible.production.cfg')
+
+    inventory_path = os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, 'inventory')
+    ansible_playbook = "ansible-playbook -i {0} {1} {2}".format(inventory_path,
```
ekohl: Also note that ```subprocess.Popen``` accepts a list which gives you shell quoting for free providing better security.

johnpmitsch:
yee-haw

johnpmitsch:
@ekohl @evgeni this has been updated according to your comments, please take another look and double check I addressed all of your concerns

bin/satellite-clone.py
johnpmitsch: I can't figure out how to do real-time output w/o the code block below, feel free to comment if you figure out a better way.

bin/satellite-clone.py
```diff
@@ -0,0 +1,68 @@
...
+def yes_or_no(question):
+    while "the answer is invalid":
+        reply = str(raw_input(question+' (y/n): ')).lower().strip()
+        if reply[0] == 'y':
+            return True
+        if reply[0] == 'n':
+            exit(1)
+
+def main():
+    parser = argparse.ArgumentParser(usage="Usage: satellite-clone [options]\n" \
```
evgeni: This currently renders as:
```
usage: Usage: satellite-clone [options]
Example: satellite-clone --start-at-task="Clean yum info"
```

If you don't pass `usage`, `argparse` autogenerates the following:
```
usage: satellite-clone.py [-h] [-v] [--start-at-task START_AT_TASK]
```

I don't really think the example is adding much here, and if we drop that, the autogenerated usage is just fine, IMHO.

evgeni:
One thing that came to my mind. Did you try calling Ansible without `subprocess`?

```python
from ansible.cli.playbook import PlaybookCLI
cli = PlaybookCLI(['ansible-playbook', ..])
cli.parse()
cli.run()
```

(thanks to @akofink for the idea)

bin/satellite-clone.py
```diff
@@ -0,0 +1,68 @@
...
+                      action="store", # optional because action defaults to "store"
+                      dest="start_at_task",
+                      help="Start at a specific task",)
+    args = parser.parse_args()
+
+
+    DEFAULT_PRODUCTION_INSTALL_PATH = "/usr/share/satellite-clone"
+    DEFAULT_PLAYBOOK_FILE = os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, "satellite-clone-playbook.yml")
+
+    yes_or_no("This will initiate satellite-clone. Do you want to proceed?")
```
ekohl: Would it make sense to have a ```--yes``` argument so it's no longer interactive?

bin/satellite-clone.py
```diff
@@ -0,0 +1,68 @@
...
+         sys.stdout.write("It looks like satellite-clone has not been installed properly, " \
+                          "{0} does not exist.".format(DEFAULT_PRODUCTION_INSTALL_PATH))
+         exit(1)
+
+    os.environ["ANSIBLE_CONFIG"] = os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, 'ansible.production.cfg')
+
+    inventory_path = os.path.join(DEFAULT_PRODUCTION_INSTALL_PATH, 'inventory')
+    ansible_playbook = ["ansible-playbook", "-i", inventory_path] + ansible_args + [DEFAULT_PLAYBOOK_FILE]
+ 
+    sys.stdout.write(" ".join(ansible_playbook))
```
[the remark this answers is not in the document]
johnpmitsch: print adds a newline so I switched to `sys.stdout.write` for the output below, and I just switched everything to use it.

bin/satellite-clone.py
johnpmitsch: I thought about this, but satellite-clone is a pretty deliberate command to run and will heavily modify a system and overwrite data. A user also only has to run it once (or maybe a few) times. Even having `satellite-clone -y` in the history or as a misunderstood option makes me nervous about wiping someone's Satellite instance, not sure if I'm being overly paranoid, but the risk/reward doesn't seem worth it.

ekohl: That's fair.

bin/satellite-clone.py
ekohl: I think it has to do with python buffering the output from subprocess by default. You can try to use ```buffer=0``` to force unbuffered output.

johnpmitsch:
@evgeni 

I updated to according to your comments, thanks again for the feedback.

There is one thing that I can't figure out though... It doesn't seem like the ansible.production.cfg is being recognized anymore:

When I run the executable, I get
```
[root@sat-628-rhel7 ~]# satellite-clone 
This will initiate satellite-clone. Do you want to proceed? (y/n): y
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions. This feature will be removed 
in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is discouraged. The module documentation details page may explain more about this rationale.. This feature will 
be removed in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```
deprecation warnings are printing out even though the environment variable ANSIBLE_CONFIG is being set and is recognized by the playbook:
```
- debug:
    msg: "{{ lookup('env', 'ANSIBLE_CONFIG') }}"

# results in 
TASK [satellite-clone : debug] ***************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "/usr/share/satellite-clone/ansible.production.cfg"
}

```

and in the config file
```
[root@sat-628-rhel7 ~]# cat /usr/share/satellite-clone/ansible.production.cfg
[defaults]
callback_plugins = /usr/share/satellite-clone/callback_plugins/
callback_whitelist = profile_tasks
inventory = /usr/share/satellite-clone/inventory
log_path = /var/log/satellite-clone/playbook.log
gathering = smart
fact_caching = jsonfile
fact_caching_connection = /usr/share/satellite-clone/cache
fact_caching_timeout = 259200
deprecation_warnings = False
```
deprecation_warnings are false but still printing out, this was working before w/ subprocess (ansible.production.cfg was recognized). I'm not sure what is different with `PlaybookCLI`.  Any ideas why this is happening? This variable should take precedence [according to the docs](http://docs.ansible.com/ansible/latest/intro_configuration.html)

bin/satellite-clone
```diff
@@ -0,0 +1,74 @@
...
+                        dest="start_at_task",
+                        help="Start at a specific task")
+    parser.add_argument("--step",
+                        action="store_true",
+                        dest="step",
+                        default=False,
+                        help="interactive: confirm each task before running")
+    args = parser.parse_args()
+
+    PRODUCTION_INSTALL_PATH = "/usr/share/satellite-clone"
```
ekohl: You could make this a parameter ```--install-path``` which could help with development setups.

bin/satellite-clone
```diff
@@ -0,0 +1,74 @@
...
+        sys.stdout.write("It looks like satellite-clone has"
+                         " not been installed properly, {0} does not exist."
+                         .format(PRODUCTION_INSTALL_PATH))
+        exit(1)
+
+    os.environ["ANSIBLE_CONFIG"] = os.path.join(PRODUCTION_INSTALL_PATH,
+                                                'ansible.production.cfg')
+
+    inventory_path = os.path.join(PRODUCTION_INSTALL_PATH, 'inventory')
+    inventory_path = os.path.join(PRODUCTION_INSTALL_PATH, 'inventory')
```
ekohl: duplicated line

bin/satellite-clone
johnpmitsch: right now we use `ansible-playbook` to run in development mode, this script is just for the rpm

bin/satellite-clone
johnpmitsch: good catch, updated

bin/satellite-clone
ekohl: Usually I find that config options are good to have to make things portable/testable but I'll leave it up to you.

bin/satellite-clone
```diff
@@ -0,0 +1,76 @@
...
+    # We need to import here because the ANSIBLE_CONFIG environment variable is
+    # checked at import time (instead of when calling PlaybookCLI)
+    from ansible.cli.playbook import PlaybookCLI
+
+    inventory_path = os.path.join(PRODUCTION_INSTALL_PATH, 'inventory')
+    ansible_playbook = (["ansible-playbook", "-i", inventory_path] +
+                        ansible_args +
+                        [DEFAULT_PLAYBOOK_FILE])
+
+    sys.stdout.write(" ".join(ansible_playbook))
```
ekohl: Is this considered debug info or something for the user?

bin/satellite-clone
```diff
@@ -0,0 +1,76 @@
...
+    if args.verbose:
+        ansible_args.append("-vvv")
+    if args.start_at_task:
+        ansible_args.append("--start-at-task")
+        ansible_args.append(args.start_at_task)
+    if args.step:
+        ansible_args.append("--step")
+
+    if not os.path.isdir(PRODUCTION_INSTALL_PATH):
+        sys.stdout.write("It looks like satellite-clone has"
```
evgeni: You need a final `\n` here, or the shell looks funny:

```
% python /tmp/satellite-clone.py
This will initiate satellite-clone. Do you want to proceed? (y/n): y
It looks like satellite-clone has not been installed properly, /usr/share/satellite-clone does not exist.%
```

`%` is how my zsh displays a missing `\n`

bash:
```
$ python /tmp/satellite-clone.py
This will initiate satellite-clone. Do you want to proceed? (y/n): y
It looks like satellite-clone has not been installed properly, /usr/share/satellite-clone does not exist.[egolov@kangae ~]$ 
```

ekohl: That inspires me to give even more comments: move this check to before the ```yes_or_no```. I can imagine it's annoying to answer a question and then fail. I do think it's good to do it after the argument parsing because then you can use ```--help```.

johnpmitsch:
@ekohl @evgeni thanks all, updated and testing on my end looks good, will run jenkins soon

johnpmitsch:
yee-haw

PR #247, opened 2017-10-25
Add example of backup dir to doc

pipopopo:
Add example of backup dir to doc


johnpmitsch:
I'm not sure about this as `katello-backup` doesn't result in a tarball, but rather many files, so that would be up to the user on how to get those files onto their machine. Was there something specific that was confusing?

pipopopo:
My bad, I'm always copying katello-backup.tar to the clone machine and forgot that original katello-backup is not compressed.

Thanks

PR #251, opened 2017-11-01
Don't use pause module

### LINKED ISSUE #250: don't use pause module
the pause module displays a prompt:
`(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort) `

This could break out of the executable script in production installs. We should use an alternative command/module that doesn't display this prompt

johnpmitsch:
Fixes #250
The pause module displays a prompt:
(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort)

This could break out of the executable script in production installs.
We should use an alternative command/module that doesn't display this prompt

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -153,8 +153,10 @@
   - name: Restart katello-service
     command: katello-service start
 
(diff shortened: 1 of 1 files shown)


johnpmitsch:
yee-haw

PR #262, opened 2017-11-13
Migrate cp database after install

### LINKED ISSUE #261: Candlepin will fail after install
After the first installation (before upgrade), candlepin will fail and this error can be found in the logs:
`Caused by: org.hibernate.HibernateException: Missing column: quantity_consumed in public.cp_pool`
the cp database has to be migrated

johnpmitsch:
Fixes #261 - The candlepin database can be restored from an
earlier sat version, so we can migrate it before running the upgrade
to avoid issues.

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -178,6 +178,17 @@
   - name: Wait for foreman-tasks service to start
     command: sleep 300
 
+  - name: Get candlepin credentials
+    command: sed -n -e 's/^.*myDS.password=//p' /etc/candlepin/candlepin.conf
+    register: candlepin_password
+
+  - set_fact:
+      clone_candlepin_password: "{{ candlepin_password.stdout }}"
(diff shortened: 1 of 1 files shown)


johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
What has been happening to cause the issues:
- katello-backup is from older 6.2.z version
- that backup is restored, cp database schema is from older 6.2.z satellite version
- cp tries to start, newer cp is installed (since newer 6.2.z is installed)
- cp can't find the columns added in new 6.2.z sat

johnpmitsch:
FYI we already [do this for pulp](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/tasks/restore.yml#L29-L30)

sthirugn:
ACK pending dolly results

johnpmitsch:
dolly passed!

PR #266, opened 2017-11-16
Actually drop mongodb

### LINKED ISSUE #259: migrating Sat6 from RHEL6 causes Sync Plans are run twice
(copy from https://bugzilla.redhat.com/show_bug.cgi?id=1511362 ):

Description of problem:
When one clones a Satellite6 from RHEL6 via satellite-clone, any Sync Plan (either cloned from a parent Satellite or a new one created) is attempted to execute twice on every schedule.

This problem does _not_ happen when cloning Sat6 on RHEL7. Only when cloning from RHEL6 (where katello-backup --online-backup was run)

The root cause is the following:
- satellite-clone installs Satellite incl. filling event_listeners mongo collection
- this collection is not dropped (cf. [1]) while I guess it should be
- once the satellite-clone finishes, event_listeners collection in mongo has 2 records (original from installer and one copied from cloned Sat)
- how Sync Plan is triggered:
  - pulp checks scheduled_calls collection to trigger repo synces
  - once a repo sync finishes, hooks from event_listeners are run to notify katello/foreman
  - so katello gets such trigger and files foreman-task for repo sync (with pulp task ID such that the foreman-task wont trigger new pulp sync task, but just checks it status)
  - pulp executes the triggers for all applicable event_listeners, so having a duplicate ones, a duplicate foreman-task is created (for each and every repo synced)

[1] https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/tasks/reset_pulp_data.yml#L2

Version-Release number of selected component (if applicable):
satellite-clone-1.1.4-1.el7sat.noarch

How reproducible:
100%

Steps to Reproduce:
1. Clone whatever Satellite (even freshly installed)
2. Have (cloned or created) Sync Plan (idealy with hourly frequency to see it soon)
3. Wait for the Sync Plan execution time and check foreman tasks created
4. check event listeners:
mongo pulp_database --eval "DBQuery.shellBatchSize = 10000000; db.event_listeners.find().shellPrint()"

Actual results:
3. Shows duplicate tasks created / half of them failing with "acquired lock .." error.
4. shows 2 listeners

Expected results:
3. no extra tasks created
4. to show just 1 listener

Additional info:
I was monitoring content of the events_listener table. The table got one new record, then:
- when cloning from RHEL7, this record was removed and replaced by the same record from original Satellite
- when cloning from RHEL6, the record from original Satellite was just _added_ but previous record remained there - thats the problem

(just out of curiosity (if I am right with the way of fix): why we dont drop *all* tables but just some of them?)

johnpmitsch:
The echo command we were using wasn't actually dropping the mongo
database, but rather echoing 'db.dropDatabase();' | mongo pulp_database

This resulted in data being restored on an existing database.

So the mongo database stayed during restore. We didn't notice many issues
because not much is in the mongo database on a clean satellite install.

Fixes #259

### DIFF
--- a/roles/satellite-clone/tasks/restore_mongo_dump.yml
+++ b/roles/satellite-clone/tasks/restore_mongo_dump.yml
@@ -1,5 +1,5 @@
 - name: Drop old mongo database
-  command: echo 'db.dropDatabase();' | mongo pulp_database
+  command: mongo pulp_database --eval "db.dropDatabase()"
 
 - name: Restore new mongo data
   command: mongorestore --host localhost {{ backup_dir }}/mongo_dump/pulp_database/
--- a/roles/satellite-clone/tasks/reset_pulp_data.yml
+++ b/roles/satellite-clone/tasks/reset_pulp_data.yml
@@ -2,7 +2,7 @@
   command: mongo pulp_database --eval "db.units_distribution.drop();db.units_docker_blob.drop();db.units_docker_image.drop();db.units_docker_manifest.drop();db.units_docker_tag.drop();db.units_drpm.drop();db.units_erratum.drop();db.units_iso.drop();db.units_package_category.drop();db.units_package_environment.drop();db.units_package_group.drop();db.units_puppet_module.drop();db.units_rpm.drop();db.units_srpm.drop();db.units_yum_repo_metadata_file.drop();db.repo_content_units.drop();"
 
 - name: reset pulp repo importers
-  command: 'echo "db.repo_importers.update({}, {\$set: {\"scratchpad.repomd_revision\": null}}, {multi: true});" | mongo pulp_database'
+  command: 'mongo pulp_database --eval "db.repo_importers.update({\"scratchpad.repomd_revision\": {$exists: true}}, {$set: {\"scratchpad.repomd_revision\": null}}, {multi: true});"'
 
 - name: reset pulp data in 6.2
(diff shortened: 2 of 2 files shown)


johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

roles/satellite-clone/tasks/restore_mongo_dump.yml
```diff
@@ -1,5 +1,5 @@
 - name: Drop old mongo database
-  command: echo 'db.dropDatabase();' | mongo pulp_database
+  command: mongo pulp_database --eval "db.dropDatabase()"
```
bbuckingham: I see a similar call in satellite-clone/tasks/reset_pulp_data.yml, should it also be updated?

johnpmitsch: Thanks brad, I updated that line as well, its possible that that wasn't executing properly as well so good catch.

Since that runs only on backups w/o pulp_data I'll add that scenario to jenkins and re-run.

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
currently jenkins is blocked by https://github.com/RedHatSatellite/satellite-clone/issues/275

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

PR #273, opened 2017-11-21
Revert "Disable REX on install"

### LINKED ISSUE #272: Don't disable REX
Disabling REX can cause errors as seen in the BZ below. There is really no clean way to remove a plugin in Satellite. We actually can keep REX enabled and not worry about talking to existing clients as ssh keys are regenerated on install so are invalidated for existing clients.

https://bugzilla.redhat.com/show_bug.cgi?id=1514383

johnpmitsch:
This reverts commit 09387bf5bb39dc179c854364f984c41192d307c7.

Fixes #272 - see issue for more context


sthirugn:
ACK

PR #278, opened 2017-11-28
Ensure satellite isn't already installed

tbrisker:


sthirugn:
currently if there are any errors in playbook, we simply rerun it.  But this new step will prevent rerunning of the playbook.  Can we have an option to override this check may be?

johnpmitsch:
Thanks for the PR!

Can you do a couple things:
- create an github issue and put "Fixes #<number>" in the commit msg
- use [fact caching](https://github.com/RedHatSatellite/satellite-clone#fact-caching) for the `satellite_installed` variable

johnpmitsch:
The pre-install check can be skipped completely, which makes things a little complicated, we would be adding the ability to skip something that can already be skipped.

I would be in favor of adding this to the [main part of the role](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/tasks/main.yml) in the beginning, after pre-install check, this way it can be skipped independent of the pre-install check.

Then you would need a way to skip, so some default variable such as `skip_satellite_rpm_check` and add `when: skip_satellite_rpm_check` to the tasks.

Then create a default False value for this variable [here]( https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/defaults/main.yml)
and add it to the [config file](https://github.com/RedHatSatellite/satellite-clone/blob/master/satellite-clone-vars.sample.yml)

let me know if you have any questions/concerns!

johnpmitsch:
@tbrisker looks good at a glance, I'll test out manually. 

It still would be nice to have an option to skip this check as the cache does clear itself (I think it persists for a week) and this check will be performed again with no way of skipping

johnpmitsch:
@tbrisker I'm thinking we have a 'skip_satellite_rpm_check" variable that defaults to false and can be overwritten in the config.
I'm open to other suggestions, the important things are: the user knows what variable to set w/o reading code, and the steps can be skipped if the cache is cleared

tbrisker:
updated

johnpmitsch:
yee-haw

johnpmitsch:
Currently failing with
```
TASK [satellite-clone : Check if satellite is already installed] ***************
Wednesday 29 November 2017  18:25:51 +0000 (0:00:00.014)       0:00:01.372 **** 
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["rpm", "-q", "satellite"], "delta": "0:00:00.280178", "end": "2017-11-29 18:25:51.966258", "failed": true, "msg": "non-zero return code", "rc": 1, "start": "2017-11-29 18:25:51.686080", "stderr": "", "stderr_lines": [], "stdout": "package satellite is not installed", "stdout_lines": ["package satellite is not installed"]}

cmd: rpm -q satellite

start: 2017-11-29 18:25:51.686080

end: 2017-11-29 18:25:51.966258

delta: 0:00:00.280178

msg: non-zero return code

stdout: package satellite is not installed
```

You need to ignore this error by [adding `ignore_errors: yes`](http://docs.ansible.com/ansible/latest/playbooks_error_handling.html#ignoring-failed-commands) or an alternative solution

tbrisker:
@johnpmitsch thanks, updated

johnpmitsch:
yee-haw

johnpmitsch:
@tbrisker 
works great! playbook ran fine the first time and the second time with no modifications, even though satellite was installed. When I deleted the cache, I encountered the proper "satellite is installed" error and needed to add "skip_satellite_rpm_check: false" to the config file as expected.

One small change, can you add this sentence to the satellite is not installed error message?
 If you would like to force a skip of this check, please set `skip_satellite_rpm_check` to false in satellite-clone-vars.yml

tbrisker:
@johnpmitsch thanks, updated!

johnpmitsch:
thanks @tbrisker!

PR #282, opened 2017-11-28
Allow multi-word activation key names

ehelms:
This allows activation keys like 'Satellite 6.3 RHEL7'


johnpmitsch:
@ehelms thanks!

Issue #292, opened 2017-12-18
trouble with cache / add --cleancache option?


pondrejk:
Suppose this scenario: my playbook run fails on `[satellite-clone : Check if root has enough space] `, which is accurate. I rearrange my partitions and run again, getting the same error. This is because  set_fact for root size is cached [1], so the next run doesn't check the actual root size. Not sure if cacheable can be dropped here, but if not, we should at least inform user about the necessity to clean the cache and where to find it (I suppose /usr/share/satellite-clone/cache/localhost). Even nicer would be to have a `satellite-clone --clean-cache` option

[1] https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/tasks/pre_install_check.yml#L28

johnpmitsch:
closing as this has been fixed, thanks @pondrejk!

PR #293, opened 2017-12-18
added --flush-cache

pondrejk:
Related to #292 Tested with intentionally corrupted cache:

```
~]# satellite-clone
This will initiate satellite-clone. Do you want to proceed? (y/n): y
ansible-playbook -i /usr/share/satellite-clone/inventory /usr/share/satellite-clone/satellite-clone-playbook.yml

PLAY [satellite6] **************************************************************
 [WARNING]: error in 'jsonfile' cache plugin while trying to read /usr/share
/satellite-clone/cache/localhost : Expecting property name enclosed in double
quotes: line 229 column 17 (char 7714). Most likely a corrupt file, so erasing
and failing.

...
ansible.errors.AnsibleError: Error while decoding the cache file /usr/share/satellite-clone/cache/localhost: The cache file /usr/share/satellite-clone/cache/localhost was corrupt, or did not otherwise contain valid data. It has been removed, so you can re-run your command now.
```

and 

```
~]# satellite-clone --flush-cache 
This will initiate satellite-clone. Do you want to proceed? (y/n): y
ansible-playbook -i /usr/share/satellite-clone/inventory --flush-cache /usr/share/satellite-clone/satellite-clone-playbook.yml

PLAY [satellite6] **************************************************************

TASK [Gathering Facts] *********************************************************
Monday 18 December 2017  16:28:18 -0500 (0:00:00.059)       0:00:00.059 ******* 
ok: [localhost]

TASK [satellite-clone : Fail if the destination host is not rhel7 (for migration only)] ***
Monday 18 December 2017  16:28:19 -0500 (0:00:00.828)       0:00:00.888 ******* 
skipping: [localhost]
...
```


johnpmitsch:
@pondrejk don't forget to add to the man page!

pondrejk:
@johnpmitsch added, thanks!

bin/satellite-clone
```diff
@@ -60,6 +66,9 @@ def main():
         ansible_args = []
         ansible_args.append("--list-tasks")
         confirm = True
+    if args.flush_cache:
+        ansible_args = []
```
johnpmitsch: This will overwrite any previous args, above we do that for `--list-tasks` since that is an independent task and won't run the playbook. If the playbook still runs when running `--flush-cache`, we should keep any other options (i.e. --verbose). You can remove this line if my above assumptions are correct

pondrejk: Didn't realize that, updated, thanks

johnpmitsch:
Thanks @pondrejk!

PR #300, opened 2018-01-15
Made portal registration disabled by default.

### LINKED ISSUE #285: Make portal registration disabled by default
Seeing how users use satellite-clone (registering to the portal and manually attaching subscription pools), it doesn't make sense to keep the default registering to portal enabled.

Recommended changes:
Make register/subscribe as a manual step before cloning.  We can still provide an option to use an activation key and password to subscribe to the portal.

In satellite-clone-vars.sample.yml:
#activationkey: changeme
#org: changeme
#register_to_portal: false

satellite-clone/defaults/main.yml:
register_to_portal: False

This is a minor change but will improve user experience greatly.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1519515

sthirugn:
By default,
- the server will not be registered to Red Hat Portal
- It is the responsibility of the user to provide right repos for
satellite installation.

Override this setting to `yes` and update `activationkey` and `org`
variables to do the portal regisration automatically.

Fixes #285

### DIFF
--- a/satellite-clone-vars.sample.yml
+++ b/satellite-clone-vars.sample.yml
@@ -1,22 +1,4 @@
 ---
-############## required variables ####################
-# Registration info - Credentials to register your Satellite to Red Hat portal
-# - You will need an activation key with a Satellite subscription.
-# - For security reasons only Activationkey / Org combination is supported for now.
-# - If you don't know how to create Activation keys in https://access.redhat.com, refer
-#   to the knowledge base article https://access.redhat.com/articles/1378093.  In this
-#   document, there are separate sections for `Creating an Activation Key` and `Finding
-#   your Organization ID`.
-# - If you want to register your Destination node to the portal by yourself, you can
-#   do that and skip the registration part in the playbook run by setting `register_to_portal`
-#   to false in this config.
-# - `org` field below is the unique Red Hat ID for your organization in portal. It may
-#   look like `1111111`.
-activationkey: changeme
-org: changeme
-
-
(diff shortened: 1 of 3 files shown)


johnpmitsch:
yee-haw

satellite-clone-vars.sample.yml
```diff
@@ -51,11 +33,24 @@ org: changeme
...
 #   - the host will not be registered to RH portal.
-#   - it is the responsibility of the user to make the required repos available.  Otherwise, the script will stop with an error.
-# uncomment the following line to override this option (defaults to true)
-#register_to_portal: true
+#   - it is the responsibility of the user to make the required repos available.
+# uncomment the following line to override this option (defaults to false). If you override this option, you must update
+# `activationkey` and `org` variables in the next section.
+#register_to_portal: false
+
+# Red Hat Portal Registration info - Credentials to register your Satellite to Red Hat portal
```
johnpmitsch: I think it would be a good idea to reinforce that you only need this section if "register_to_portal" is true here

johnpmitsch:
yee-haw

satellite-clone-vars.sample.yml
sthirugn: sure, added ```Use this section if and only if `register_to_portal` is `true`.```

johnpmitsch:
yee-haw

johnpmitsch:
Thanks @sthirugn!

PR #316, opened 2018-01-31
Steps to restore pulp content after cloning.

### LINKED ISSUE #298: [Need Documentation] Will cloning process work to use a backup with the “--skip-pulp-content” option, or do we need a backup with the full amount of content?
you can use --skip-pulp-content but it takes a bunch of work to re-setup all the content because you have to:

1) flip all yum repos to on-demand:
```sh
# DOWNLOAD_POLICY=on_demand foreman-rake  katello:change_download_policy
```
2) re-sync every repo
3) republish every content view

Note: You also lose anything you uploaded via the UI/hammer

sthirugn:
Fixes #298

### DIFF
--- a/docs/satellite-clone.md
+++ b/docs/satellite-clone.md
@@ -3,12 +3,10 @@
 #### Important Notes ####
 Before cloning:
   - It is recommended that the target server is network isolated. We take steps to disable communication from the target server to existing capsules and hosts, but the only way to ensure this is in an isolated environment.
-  - If your storage is slow (like NFS) and your pulp backup tar file is large (>100 gb), you might see memory errors while untaring pulp data during the playbook run. In this case don't include pulp_data.tar during the cloning process and instead copy over `/var/lib/pulp` from the source server to the target server after the cloning process. For example, from the source server:
-
-     ```console
-        # rsync -az /var/lib/pulp target_server:/var/lib/pulp
-    ```
-    Note that rsync command may take a while to complete to copy large data.
+  - You can choose to perform a clone without a pulp_data.tar file. After the clone run *without pulp content*, you have two options:
+    - To recover pulp content after the cloning process, follow these [steps](#cloning-without-pulp_contenttar-and-recover-pulp-content-after-cloning).
+    - If you are cloning just for testing purposes, you can get away with not using pulp data at all. The target server should still be functional other than the pulp content.
+  - If your storage is slow (like NFS) and your pulp backup tar file is large (>100 gb), you might see memory errors while untaring pulp data during the playbook run. In this case, don't include pulp_data.tar during the cloning process.
 
 During cloning:
   - The playbook will update the target server's hostname to match the hostname of the source server.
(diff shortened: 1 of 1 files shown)


docs/satellite-clone.md
```diff
@@ -3,12 +3,9 @@
...
 Before cloning:
   - It is recommended that the target server is network isolated. We take steps to disable communication from the target server to existing capsules and hosts, but the only way to ensure this is in an isolated environment.
-  - If your storage is slow (like NFS) and your pulp backup tar file is large (>100 gb), you might see memory errors while untaring pulp data during the playbook run. In this case don't include pulp_data.tar during the cloning process and instead copy over `/var/lib/pulp` from the source server to the target server after the cloning process. For example, from the source server:
-
-     ```console
-        # rsync -az /var/lib/pulp target_server:/var/lib/pulp
-    ```
-    Note that rsync command may take a while to complete to copy large data.
+  - If your storage is slow (like NFS) and your pulp backup tar file is large (>100 gb), you might see memory errors while untaring pulp data during the playbook run. In this case, don't include pulp_data.tar during the cloning process.  After the clone run, you have two options:
+    - To recover pulp content after the cloning process, follow these [steps](#recover-pulp-content-after-cloning).
```
johnpmitsch: I think we should make a clear distinction that this is an option only when you clone without pulp_data.tar

docs/satellite-clone.md
```diff
@@ -3,12 +3,9 @@
...
-  - If your storage is slow (like NFS) and your pulp backup tar file is large (>100 gb), you might see memory errors while untaring pulp data during the playbook run. In this case don't include pulp_data.tar during the cloning process and instead copy over `/var/lib/pulp` from the source server to the target server after the cloning process. For example, from the source server:
-
-     ```console
-        # rsync -az /var/lib/pulp target_server:/var/lib/pulp
-    ```
-    Note that rsync command may take a while to complete to copy large data.
+  - If your storage is slow (like NFS) and your pulp backup tar file is large (>100 gb), you might see memory errors while untaring pulp data during the playbook run. In this case, don't include pulp_data.tar during the cloning process.  After the clone run *without pulp_data.tar*, you have two options:
+    - To recover pulp content after the cloning process, follow these [steps](#cloning-without-pulp_contenttar-and-recover-pulp-content-after-cloning).
```
johnpmitsch: We should specify the pulp_data part here, its still not clear that this step isn't required when you clone w/ pulp_data.tar when you first read this.

sthirugn: the previous line states - `After the clone run *without pulp_data.tar*, you have two options:`. do you have recommendations on how you want this line to be?

johnpmitsch: ok I see now, I missed that part, it looks much more obvious when I look at it formatted.

One thing is I'm not crazy about making assumptions and giving users the If this doesn't work, then do this phrasing. Can you have a separate line under the memory error part for the second sentence that says something along the lines of: `If you would like to use a backup without a pulp_data.tar file, you have these two options`. I'm guessing a lot of users will want to do this from the start (regardless of memory issues).

sthirugn:
@johnpmitsch updated

PR #318, opened 2018-02-01
Updated satellite rpm check

sthirugn:
- Renamed the variable `skip_satellite_rpm_check` to `disable_satellite_rpm_check` to make it consistent with other variables.
- Added documentation in satellite-clone-vars.sample file about the clearing cache option to force run this task.

### DIFF
--- a/roles/satellite-clone/tasks/pre_install_check.yml
+++ b/roles/satellite-clone/tasks/pre_install_check.yml
@@ -9,14 +9,14 @@
   register: satellite_rpm
   ignore_errors: yes
   when:
-    - not skip_satellite_rpm_check
+    - not disable_satellite_rpm_check
     - clone_satellite_installed_check is not defined
 
 - name: Fail if satellite is already installed
   fail:
     msg: "Satellite is already installed. Satellite-clone should be run on a clean RHEL7 machine. If you would like to force a skip of this check, please set skip_satellite_rpm_check to true in satellite-clone-vars.yml"
   when:
-    - not skip_satellite_rpm_check
+    - not disable_satellite_rpm_check
     - clone_satellite_installed_check is not defined
     - satellite_rpm.rc | int == 0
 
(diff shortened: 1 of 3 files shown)


satellite-clone-vars.sample.yml
```diff
@@ -65,7 +65,9 @@
 #restorecon: false
 
 # Disable checking that satellite is already installed. (defaults to false)
-#skip_satellite_rpm_check: false
+# Once the playbook confirms that there is no satellite installation, the subsequent runs will not run this task.
+# To force run this task for a specific run, use `--flush-cache` option
```
johnpmitsch: I'm fine with changing the variable name for consistency, but I think this explanation is overly confusing from a user perspective. The intent behind having the check only run on the first run is so the user doesn't have to worry about it. If they do hit the satellite rpm check unnecessarily, the error message is pretty clear about what to do to work around the blockage. I view this as some complexity a user shouldn't have to deal with unless they have an issue with it.

That being said, if you want to add comments to the code for our own sanity, I'm all for that :smile:

roles/satellite-clone/tasks/pre_install_check.yml
```diff
@@ -9,14 +9,14 @@
   register: satellite_rpm
   ignore_errors: yes
   when:
-    - not skip_satellite_rpm_check
+    - not disable_satellite_rpm_check
     - clone_satellite_installed_check is not defined
 
 - name: Fail if satellite is already installed
   fail:
     msg: "Satellite is already installed. Satellite-clone should be run on a clean RHEL7 machine. If you would like to force a skip of this check, please set skip_satellite_rpm_check to true in satellite-clone-vars.yml"
```
johnpmitsch: You will have to change the variable name in the error message as well

johnpmitsch:
@sthirugn do you plan on updating this? Can we close it if not?

johnpmitsch:
@sthirugn closing since this hasn't been active for months, please reopen if you would like to continue to work on this PR

sthirugn:
Thank you for closing this out, not needed for now.

PR #323, opened 2018-02-07
Remove verbose mode

johnpmitsch:
This isn't working in production, it looks like there is an
issue with PlaybookCLI API, until we figure that out,
we should remove the option since it does nothing.


johnpmitsch:
New menu:
```
[root@sat-snap-rhel7 ~]# satellite-clone --help
usage: satellite-clone [-h] [--start-at-task START_AT_TASK] [--step]
                       [--list-tasks] [--flush-cache]

optional arguments:
  -h, --help            show this help message and exit
  --start-at-task START_AT_TASK
                        Start at a specific task
  --step                interactive: confirm each task before running
  --list-tasks          list tasks that will be run in the playbook
  --flush-cache         clear the fact cache and run playbook
```

sthirugn:
ACK pending tests

sthirugn:
Can you fix the man pages too:

man/satellite-clone.md
man/satellite-clone.1

sthirugn:
ACK

PR #325, opened 2018-02-26
Add 6.3 support

johnpmitsch:
This adds an ansible module to check the metadata file in the backup
for a specific version. Because metadata.yml was added in 6.2.z,
the previous check of satellite-answers is kept for older 6.2 backups.
Any backup with metadata.yml will use the metadata file as a check.

Todo:

- [x] test on 6.3 normal backup
- [x] test on 6.3 online backup
- [x] test on 6.2 backup with metadata.yml

### DIFF
--- a/library/check_metadata_for_satellite_version.py
+++ b/library/check_metadata_for_satellite_version.py
@@ -1,0 +1,50 @@
+import yaml
+import re
+from ansible.module_utils.basic import *
+
+
+# module: check_metadata_for_satellite_version
+# description:
+#    - Return the Satellite version specified in a Satellite backup
+# notes:
+#    - The Satellite version is determined from the Satellite rpm
+#      version using the backup's rpm list from metadata.yml
+# options:
+#    metadata_path:
+#        description:
+#          - Full path (including file name) to metadata.yml
+#        required: true
+def get_satellite_backup_version(params):
+    with open(params["metadata_path"]) as data_file:
+        data = yaml.load(data_file)
+
+    rpms = data[":rpms"]
+    pattern = re.compile("^satellite-[\d+].*")
+    satellite = [r for r in rpms if pattern.match(r)][0]
+    satellite_version = satellite.split("-")[1]
+
+    found_version = '.'.join(satellite_version.split('.')[0:2])
+
+    if found_version not in ["6.2", "6.3"]:
+        return False, dict(msg="Satellite Version is not supported")
+
+    msg = "{0} backup found".format(found_version)
+    result = dict(satellite_version=found_version, msg=msg, changed=False)
+    return True, result
+
+
+def main():
+    fields = {
+        "metadata_path": {"required": True, "type": "str"}
+    }
+
+    module = AnsibleModule(argument_spec=fields)
+    success, result = get_satellite_backup_version(module.params)
+    if success:
+        module.exit_json(**result)
+    else:
+        module.fail_json(**result)
+
+
+if __name__ == '__main__':
+    main()
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -2,7 +2,7 @@
 username = "admin"
 password = "changeme"
 external_capsules = []
-external_capsule_ids = `hammer -u #{username} -p #{password} --csv capsule list --search 'feature = \"Pulp Node\"' | tail -n+2 | awk -F, {'print $1'}`
+external_capsule_ids = `hammer -u #{username} -p #{password} --csv capsule list --search 'feature = \"Pulp Node\"' | grep -v "Warning:" | tail -n+2 | awk -F, {'print $1'}`
 if external_capsule_ids.empty?
   STDOUT.puts "There are no external capsules to disassociate."
 else
--- a/roles/satellite-clone/tasks/backup_satellite_version_check.yml
+++ b/roles/satellite-clone/tasks/backup_satellite_version_check.yml
@@ -1,0 +1,37 @@
+- name: check for satellite answers file
+  command: tar zxf {{ backup_dir }}/config_files.tar.gz  etc/foreman-installer/scenarios.d/satellite-answers.yaml --to-stdout
+  register: satellite_answers
+  ignore_errors: true
+  no_log: true
+
+- name: setting fact - satellite_answers
+  set_fact:
+    clone_satellite_answers: "{{ (satellite_answers.rc | int) == 0 }}"
+    cacheable: true
+
+- name: Temporarily set satellite_version to 6.2 if satellite_answers.yaml exists
+  set_fact:
+    satellite_version: "6.2"
+    cacheable: true
+  when: clone_satellite_answers
+
+- name: set satellite_version to 6.1 if satellite_answers.yaml does not exist
+  set_fact:
+    satellite_version: "6.1"
+    cacheable: true
+  when: not clone_satellite_answers
+
+- name: Check for satellite version in metadata file
+  check_metadata_for_satellite_version:
+    metadata_path: "{{ backup_dir }}/metadata.yml"
+  register: metadata_satellite_version
+  when: clone_metadata_exists
+
+- name: setting fact - satellite_version
+  set_fact:
+    satellite_version: "{{ metadata_satellite_version.satellite_version }}"
+    cacheable: true
+  when:
+    - clone_metadata_exists
+    - metadata_satellite_version is defined
+    - metadata_satellite_version | success
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -1,31 +1,10 @@
 ---
-- include: pre_install_check.yml
+- include_tasks: pre_install_check.yml
   when: run_pre_install_check
 
-- name: check satellite version from the config files
-  command: tar zxf {{ backup_dir }}/config_files.tar.gz  etc/foreman-installer/scenarios.d/satellite-answers.yaml --to-stdout
-  register: satellite_answers
-  ignore_errors: true
-  no_log: true
+- include_tasks: backup_check.yml
 
-- name: setting fact - satellite_answers
-  set_fact:
-    clone_satellite_answers: "{{ (satellite_answers.rc | int) == 0 }}"
-    cacheable: true
-
-- name: set satellite_version to 6.2 if satellite_answers.yaml exists
-  set_fact:
-    satellite_version: 6.2
-    cacheable: true
-  when: clone_satellite_answers
-
-- name: set satellite_version to 6.1 if satellite_answers.yaml does not exist
-  set_fact:
-    satellite_version: 6.1
-    cacheable: true
-  when: not clone_satellite_answers
-
-- include: backup_check.yml
+- include_tasks: backup_satellite_version_check.yml
 
 # Identify hostname from backup config file
 - name: Identify the hostname from the backup config tar file
@@ -124,10 +103,10 @@
 # Install Satellite packages
 - name: Install Satellite 6.1 packages
   yum: name=katello state=latest
-  when: satellite_version == 6.1
-- name: Install Satellite 6.2 packages
+  when: satellite_version == "6.1" 
+- name: "Install Satellite {{ satellite_version }} packages"
   yum: name=satellite state=latest
-  when: satellite_version == 6.2
+  when: satellite_version in ["6.2", "6.3"]
 
 # The postgres user is created after installing postgresql packages, so
 # we perform this owner/group change at this point rather than earlier
@@ -175,7 +154,7 @@
 # Workaround for Issue #72 -  satellite-clone playbook fails if /etc/katello-installer isn't present.
 - name: Create /etc/katello-installer folder
   file: path=/etc/katello-installer state=directory mode=0755
-  when: satellite_version == 6.2
+  when: satellite_version in ["6.2", "6.3"]
 
 # Restore Config
 - name: untar config files (for cloning only)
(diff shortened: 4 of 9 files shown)


johnpmitsch:
yee-haw

johnpmitsch:
@evgeni If you have a minute to look at this from an ansible + python perspective, that would be awesome :smile:

evgeni:
Is there a reason you do a "check for version" (twice) instead of "get the version"? plus, you're looking at the katello version, you could also look at the satellite rpm?

evgeni:
I think I'd make the module in a way, that it gets the path to a config tarball, and unarchive, read etc all happens in the module, and the return is `[6.3, 6.2, 'unknown']`, and if it's unknown, you can assume 6.1 later.

evgeni:
oh, the unarchive is not for the metadata, me--

johnpmitsch:
@evgeni Do you think I should check both satellite rpm + katello plugin version? Maybe I should just check the satellite rpm as that is the version I need in the first place. 

My only be concern is grabbing another satellite rpm (like satellite-cli) by mistake, but I should be able to put together some regex to match :)

johnpmitsch:
> return is [6.3, 6.2, 'unknown'], and if it's unknown, you can assume 6.1 later

I like returning a string with the version instead of the t/f check for version. 6.1 backups don't have metadata.yaml files (and older 6.2 z streams) so I would probably error if the module doesn't find 6.3 or 6.2 when there is a metadata.yml file present.

evgeni:
Given this is `satellite-clone` (and not `katello-clone`), I'd just check the `satellite` RPM and ignore `katello`.

johnpmitsch:
@evgeni updated the module to check the satellite rpm if you can take a look

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -2,7 +2,7 @@
 username = "admin"
 password = "changeme"
 external_capsules = []
-external_capsule_ids = `hammer -u #{username} -p #{password} --csv capsule list --search 'feature = \"Pulp Node\"' | tail -n+2 | awk -F, {'print $1'}`
+external_capsule_ids = `hammer -u #{username} -p #{password} --csv capsule list --search 'feature = \"Pulp Node\"' | grep -v "Warning:" | tail -n+2 | awk -F, {'print $1'}`
```
sthirugn: Can you explain why this change was made? this doesn't seem to be related to this PR.

roles/satellite-clone/tasks/backup_satellite_version_check.yml
```diff
@@ -0,0 +1,37 @@
...
+  register: satellite_answers
+  ignore_errors: true
+  no_log: true
+
+- name: setting fact - satellite_answers
+  set_fact:
+    clone_satellite_answers: "{{ (satellite_answers.rc | int) == 0 }}"
+    cacheable: true
+
+- name: set satellite_version to 6.2 if detected
```
sthirugn: to match the name of the next task perhaps name this `set satellite_version to 6.2 if satellite_answers.yaml exists`

roles/satellite-clone/tasks/main.yml
```diff
@@ -190,17 +169,33 @@
...
   command: katello-installer --capsule-dns false --capsule-dhcp false --capsule-tftp false
-  when: satellite_version == 6.1
-- name: run Satellite 6.2 installer
+  when: satellite_version == "6.1"
+
+- name: "Run Satellite 6.2 installer"
   command: satellite-installer --scenario satellite --foreman-proxy-dns false --foreman-proxy-dhcp false --foreman-proxy-tftp false --foreman-ipa-authentication false
-  when: satellite_version == 6.2
+  environment:
+    HOSTNAME: "{{ hostname }}"
```
sthirugn: what is the need for this?

roles/satellite-clone/tasks/main.yml
```diff
@@ -190,17 +169,33 @@
...
+  when: satellite_version == "6.2"
+
+- name: "Run Satellite 6.3 installer"
+  command: satellite-installer --scenario satellite --foreman-proxy-dns false --foreman-proxy-dhcp false --foreman-proxy-tftp false --foreman-ipa-authentication false --disable-system-checks
```
sthirugn: - what is the purpose of --disable-system checks?
- Is this applicable only for 6.3?

roles/satellite-clone/tasks/main.yml
```diff
@@ -190,17 +169,33 @@
   command: restorecon -R /
   when: restorecon
 
-# Run Satellite installer
-- name: run Satellite 6.1 installer
+# This file tells the candlepin puppet module in the Satellite installer that
+# candlepin is already set up. We remove it so candlepin is set up correctly.
+- name: Remove cpdb_done file
```
sthirugn: did this cause an issue in the past? If so we can create an issue and link to this PR too.

roles/satellite-clone/tasks/main.yml
```diff
@@ -243,20 +240,20 @@
 
   - name: update katello assets
     file: src=/opt/rh/ruby193/root/usr/share/gems/gems/katello-2.2.0.93/public/assets/katello dest=/usr/share/foreman/public/assets/katello
-    when: satellite_version == 6.1
+    when: satellite_version == "6.1"
```
sthirugn: what is the reason behind adding double quotes here?

roles/satellite-clone/tasks/restore.yml
```diff
@@ -6,10 +6,10 @@
   service: name=postgresql state=stopped
 
 - name: Restore postgresql data
-  command: tar --selinux --overwrite -xf {{ backup_dir }}/pgsql_data.tar.gz -C /
+  command: "tar --selinux --overwrite -xf {{ backup_dir }}/pgsql_data.tar.gz -C /"
```
sthirugn: what is the reason behind including double quotes here?

roles/satellite-clone/tasks/backup_satellite_version_check.yml
johnpmitsch: Thanks, since it will be updated to 6.3 if its a 6.3 backup, how about we say `Temporarily set satellite_version to 6.2 if satellite_answers.yaml exists`

roles/satellite-clone/tasks/main.yml
johnpmitsch: So this is something that only occurs in 6.3, but I added in 6.2 as well to be safe. 

The $HOSTNAME bash variable doesn't update when running the playbook after we change the hostname. Some of the puppet modules that the installer runs rely on this bash variable. In this case it was making calls to the foreman-proxy using $HOSTNAME, but using the old $HOSTNAME bash variable, before it was changed to match the backup's hostname. This ensures that the bash variable is updated.

roles/satellite-clone/tasks/main.yml
johnpmitsch: Yes, I was confused about this as well, I guess it is not an option in 6.2 (I thought it was...) but the RAM check is going to block jenkins. I figure if someone has already installed and backed up Satellite on one machine (which they have if they are running clone), they understand the system requirements needed to run Satellite and we shouldn't be checking this as part of sat-clone and failing playbooks because of the system requirements check.

roles/satellite-clone/tasks/main.yml
johnpmitsch: I don't think it has in the past, but it definitely is causing an issue in 6.3.

roles/satellite-clone/tasks/main.yml
johnpmitsch: So the check_metadata_for_satellite_version module returns a string and I can't get it to return an integer. I'm guessing this has to do with how `register` works in ansible. I changed satellite_version to always be a string to match the new module and updated usage appropiately

roles/satellite-clone/tasks/restore.yml
johnpmitsch: It was part of debugging a different issue (that was solved by the delete `cp_done` task) but I think we actually should be using quotes wherever string interpolation `{{ string }}` is used

roles/satellite-clone/files/disassociate_capsules.rb
johnpmitsch: So on 6.3 hammer is displaying a warning for the first line of the hammer command output. The `tail` command here is expecting everything after the second line, which makes sense for how the normal hammer csv output is, but the warning is bumping everything down a line. I added this to remove any lines containing `Warning:` so the count remains correct.

sthirugn:
Testing in progress:
- [x] 6.2 clone test - PASS
- [ ] 6.3 clone test
- [ ] 6.4 clone test

johnpmitsch:
@evgeni friendly poke, can you take another glance over the code? :pray:

library/check_metadata_for_satellite_version.py
```diff
@@ -0,0 +1,55 @@
...
+def check_metatadata_for_backup(params):
+    with open(params["metadata_path"]) as data_file:
+        data = yaml.load(data_file)
+
+    rpms = data[":rpms"]
+    pattern = re.compile("^satellite-[\d+].*")
+    satellite = [r for r in rpms if pattern.match(r)][0]
+    satellite_version = satellite.split("-")[1]
+
+    if LooseVersion(satellite_version) >= LooseVersion("6.2.0") and \
```
evgeni: Is there a reason not to just do
```python
found_version = '.'.join(satellite_version.split('.')[0:1])
```

library/check_metadata_for_satellite_version.py
```diff
@@ -0,0 +1,55 @@
...
+#    - Return the Satellite version specified in a Satellite backup
+# notes:
+#    - The Satellite version is determined from the Satellite rpm
+#      version using the backup's rpm list from metadata.yml
+# options:
+#    metadata_path:
+#        description:
+#          - Full path (including file name) to metadata.yml
+#        required: true
+def check_metatadata_for_backup(params):
```
evgeni: I'd argue this is not a check, but `get_satellite_backup_version`? :)

roles/satellite-clone/tasks/main.yml
evgeni: but, given we're restoring a working backup, shouldn't candlepin actually be properly setup at this point?

roles/satellite-clone/tasks/main.yml
evgeni: I wonder if [`meta: reset_connection`](https://docs.ansible.com/ansible/latest/meta_module.html) would help here and be less of a hack

evgeni:
@johnpmitsch added some comments, nothing major

roles/satellite-clone/tasks/main.yml
```diff
@@ -229,11 +224,13 @@
...
     command: foreman-rake foreman_tasks:cleanup TASK_SEARCH='label ~ *' STATES='paused'  VERBOSE=true AFTER='0h'  VERBOSE=true
-    when: satellite_version == 6.2
+    when: satellite_version in ["6.2", "6.3"]
 
-  - name: Run installer upgrade (satellite 6.2 only)
+  - name: Run installer upgrade (satellite 6.2+ only)
     command: satellite-installer --upgrade
-    when: satellite_version == 6.2
+    environment:
+      HOSTNAME: "{{ hostname }}"
```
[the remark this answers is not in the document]
johnpmitsch: meta comment?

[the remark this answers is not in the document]
evgeni: https://github.com/RedHatSatellite/satellite-clone/pull/325#discussion_r174420139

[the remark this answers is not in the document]
johnpmitsch: missed that one, thanks!

roles/satellite-clone/tasks/main.yml
johnpmitsch: not exactly, here are the rough steps of clone and what is happening
- restore configs  <---- cpdb_done file is restored from the backup
- install Satellite <----- installer sees cpdb_done file and says "ok I wont set up candlepin then", resulting in errors on seed steps when installing
- restore backups (mongo, postgres, pulp data)
- re-run installer

johnpmitsch:
@evgeni updated

johnpmitsch:
yee-haw

sthirugn:
Update satellite-clone.md to reflect the fact that metadata.yml is mandatory for 6.3 and above

sthirugn:
updated test status in comment https://github.com/RedHatSatellite/satellite-clone/pull/325#issuecomment-372759142

johnpmitsch:
yee-haw

sthirugn:
ACK

johnpmitsch:
yee-haw

roles/satellite-clone/files/disassociate_capsules.rb
evgeni: Which warning, is there a BZ about that? :-)

roles/satellite-clone/tasks/main.yml
evgeni: Aah, gotcha. I'd argue there should be an RFE to the installer, to force Candlepin re-init (like we can with certs), but that remove will do for now.

johnpmitsch:
@evgeni good point, here is [hammer BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1556948) and [cpdb_done RFE for katello](http://projects.theforeman.org/issues/22913)

PR #330, opened 2018-04-25
cleanup Satellite version differences

sean797:
Used this today to clone a 6.2 instance and test a 6.3 upgrade, it worked really well. Just had some issues due to:
-  the backup didn't include the custom certs I'm using
   - easy enough to copy them over manually.
- this role trying chown postgres <backup dir>, which was on a read-only NFS mount.
  - worked around it by chown'ing it on the NFS server.

Anyway made some small code improvements. Thanks!

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -6,6 +6,9 @@
 
 - include_tasks: backup_satellite_version_check.yml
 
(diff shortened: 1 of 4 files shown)


johnpmitsch:
yee-haw

johnpmitsch:
Thanks @sean797!

PR #340, opened 2018-08-13
Fixes #342 - Quote lifecycle environment name

### LINKED ISSUE #342: disassociate_capsules.rb fails when names of lifecycle environment contains spaces
```
# hammer lifecycle-environment list
---|------------------|--------
ID | NAME             | PRIOR  
---|------------------|--------
3  | ACME Development | Library

# ruby  disassociate_capsules.rb 
Could not remove the lifecycle environment from the capsule:
  Error: Too many arguments.
  
  See: 'hammer capsule content remove-lifecycle-environment --help'.
All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite and any interference with existing infrastructure. To reverse these changes, run the following commands, making sure to replace the credentials with your own.
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 2 --environment ACME Development --organization "Default Organization"
```
because environment name is not quoted
https://github.com/RedHatSatellite/satellite-clone/blob/2ecf359b929035d8b7153b80619904690e3d0f96/roles/satellite-clone/files/disassociate_capsules.rb#L20-L21

alda519:
fixes errors like
```
TASK [satellite-clone : Disassociate capsules with lifecycle environments (to avoid the cloned Satellite server talking with live capsules)] ***
...
stdout: Could not remove the lifecycle environment from the capsule:
  Error: too many arguments
  
  See: 'hammer capsule content remove-lifecycle-environment --help'
...
All Capsules are unassociated with any lifecycle environments. This is to avoid ...
hammer ... --csv capsule content add-lifecycle-environment --id 25 --environment Env Dev --organization "Org"
```

### DIFF
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -17,8 +17,8 @@ else
   reverse_commands = []
   external_capsules.each do |capsule|
     capsule[:lifecycle_environments].each do |env|
-      `hammer -u #{username} -p #{password} --csv capsule content remove-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization "#{capsule[:organization]}"`
-      reverse_command = "hammer -u #{username} -p #{password} --csv capsule content add-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization \"#{capsule[:organization]}\""
+      `hammer -u #{username} -p #{password} --csv capsule content remove-lifecycle-environment --id #{capsule[:id]} --environment "#{env}" --organization "#{capsule[:organization]}"`
+      reverse_command = "hammer -u #{username} -p #{password} --csv capsule content add-lifecycle-environment --id #{capsule[:id]} --environment \"#{env}\" --organization \"#{capsule[:organization]}\""
       reverse_commands << reverse_command
     end
   end


chris1984:
@johnpmitsch can this get a merge?

chris1984:
yee-haw

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -17,8 +17,8 @@
   reverse_commands = []
   external_capsules.each do |capsule|
     capsule[:lifecycle_environments].each do |env|
-      `hammer -u #{username} -p #{password} --csv capsule content remove-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization "#{capsule[:organization]}"`
-      reverse_command = "hammer -u #{username} -p #{password} --csv capsule content add-lifecycle-environment --id #{capsule[:id]} --environment #{env} --organization \"#{capsule[:organization]}\""
+      `hammer -u #{username} -p #{password} --csv capsule content remove-lifecycle-environment --id #{capsule[:id]} --environment "#{env}" --organization "#{capsule[:organization]}"`
+      reverse_command = "hammer -u #{username} -p #{password} --csv capsule content add-lifecycle-environment --id #{capsule[:id]} --environment \"#{env}\" --organization \"#{capsule[:organization]}\""
```
johnpmitsch: why are the quotes escaped here and not in the line above?

johnpmitsch:
@alda519 generally we ask you file [an issue detailing the problem](https://github.com/RedHatSatellite/satellite-clone#development) first and add `Fixes #XX` to the commit message with the issue number. You can file an issue [here](https://github.com/RedHatSatellite/satellite-clone/issues).

I see the error, but I'm confused what is different with your setup that makes you hit this error. Do you have an idea of why it fails on this step? Any more info would be appreciated

roles/satellite-clone/files/disassociate_capsules.rb
johnpmitsch: ah nvm, I see that is the pattern used with `capsule[:organization]` before, I'm not sure why we do this, I'll have to look more :man_shrugging:

alda519:
Ah, sorry, I looked at recently [closed pull requests](https://github.com/RedHatSatellite/satellite-clone/pulls?q=is%3Apr+is%3Aclosed) and there are none following that rule, so I thought that it is not necessary..

So there is new issue #342 and I also changed the commit message.

evgeni:
sorry to hijack this PR a bit, but what about the username or password containing spaces or shell special chars like `$`? this will result in similar issues.

also, using double quotes to quote the CV and Org names results in the shell  interpreting special chars (`$`, backticks) inside, so if your CV is called `Dollar$CV`, you end up with a wrong command line.

roles/satellite-clone/files/disassociate_capsules.rb
evgeni: the one is executed directly, the other is stored to a var. in the second case, you already have `""` for the var assignment, so you can't use literal `"` inside the var

johnpmitsch: thanks, I saw that after I made the comment :+1:

mccun934:
Tested this on a Satellite with multiple LCE's with spaces in the name and it properly removes the envs from all capsules. 

:+1: 

ACK

johnpmitsch:
merging this to unblock current users, @evgeni I'll file a separate issue around special chars in the script

johnpmitsch:
@evgeni issue filed https://github.com/RedHatSatellite/satellite-clone/issues/343

Issue #343, opened 2018-08-30
Handle special characters in disassociate capsules script


johnpmitsch:
see [this comment](https://github.com/RedHatSatellite/satellite-clone/pull/340#issuecomment-417309631) for more context

Issue #175, opened 2017-04-07
Capsule disassociation failed with regular clone scenario of rhel7 sat 6.2


sthirugn:
```sh
TASK [satellite-clone : Disassociate capsules with lifecycle environments (to avoid the cloned Satellite server talking with live capsules)] ***
Friday 07 April 2017  12:58:35 -0400 (0:00:00.026)       0:56:25.277 ********** 
changed: [localhost]

stdout: All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite and any interference with existing infrastructure. To reverse these changes, run the following commands, making sure to replace the credentials with your own.
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 2 --environment QE --organization "S?o Paulo"
hammer -u admin -p changeme --csv capsule content add-lifecycle-environment --id 3 --environment QE --organization "S?o Paulo"
```

Additional info:
```sh
hammer -u admin -p changeme organization list
---|----------------------|--------------------------------------|----------------------------
ID | NAME                 | LABEL                                | DESCRIPTION                
---|----------------------|--------------------------------------|----------------------------
1  | Default Organization | Default_Organization                 |                            
3  | São Paulo            | e9db3daa-5c32-4a41-881d-3fa3a828a5c1 | The São Paulo organization.
---|----------------------|--------------------------------------|----------------------------

johnpmitsch:
Looks like the script won't handle special characters well.
Unfortunately capsule info only returns capsule info by [name](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/files/disassociate_capsules.rb#L9)

We can maybe use --json and parse that

akofink:
While parsing json is one solution, I would be interested in where the special characters are not supported. During preliminary testing in IRB, it seems that Ruby 2.3 can parse special characters returned from hammer and pass them back to hammer as options in a subsequent command. Other things to try:

- Does the default shell on RHEL6/7 that this is being run on support the special characters/UTF-8?
- Check the Ruby version installed by default on the failing machine, and test with that specific version. Perhaps Ruby 1.9 or 2.0 doesn't support system commands with UTF-8.

If all else fails, then json parsing it is!

rplevka:
Constructing the hammer command and parsing its output may go wrong on many levels.
Why don't we use the proxy straight away and use json module to parse the response?

[...]

PR #344, opened 2018-08-31
shellescape all strings when disassociating capsules

evgeni:
Fixes: #343

### DIFF
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -1,15 +1,28 @@
 #!/usr/bin/ruby
-username = "admin"
-password = "changeme"
+
+require 'shellwords'
+
+@username = "admin"
+@password = "changeme"
+
+def prepare_hammer_cmd(command)
+  "hammer -u #{@username.shellescape} -p #{@password.shellescape} #{command}"
+end
+
+def run_hammer_cmd(command)
+  command = prepare_hammer_cmd(command)
+  `#{command}`
+end
+
 external_capsules = []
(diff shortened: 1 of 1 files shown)


johnpmitsch:
Do we need anything passed into the hammer commands escaped with shell escape? i.e. a lifecycle environment name

evgeni:
We have to escape everything that may contain chars outside of a-z0-9 (broadly speaking).

johnpmitsch:
its also worth noting that we probably haven't ran into this for username and password because the sat-clone script resets the admin password to 'changeme' and we hardcode it in the script. I still think we should escape them like you have in case we make that step configurable.

I thought this sounded familiar, I found [this issue](https://github.com/RedHatSatellite/satellite-clone/issues/175),  in which a user has a special character in the org, breaking the script for him. We should probably address this along with the username/password

Let me know if you want to address issue #175 in this PR (escaping the organization and LEs as well), if not we can address separately.

evgeni:
I'd argue #175 is an encoding issue, not an escaping one.

This PR does escape username, password, LE and org. I think that should be sufficient for now.

johnpmitsch:
@evgeni thanks, I missed the shellescapes on the org and LEs, will test soon :+1:

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

johnpmitsch:
changed one of the backups in dolly to be one with a capsule so the script runs :)

johnpmitsch:
yee-haw

johnpmitsch:
yee-haw

PR #346, opened 2018-11-13
Support for cloning 6.4 Satellite

jturel:
I was able to clone my satellite including pulp data with this patch!

There are a few places in the main.yml which references version 6.2, 6.3 but I didn't add 6.4 there arbitrarily. Things seem to be OK but would like some specific feedback on those omissions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,20 @@
 # Satellite-clone
 
-Easily setup a Satellite 6.1, 6.2, or 6.3 server with restored backup data.
+Easily set up a Red Hat Satellite server with restored backup data.
 
 ## Getting Started
 Throughout this documentation, ensure that you understand the following terminology:
 - Source server: Existing Satellite server.
 - Target server: new server, to which Satellite server is being cloned.
 
+#### Supported Satellite Versions ####
+- 6.2
+- 6.3
+- 6.4 (remote database cloning not currently supported)
+
 #### What you need: ####
   - A blank (vanilla install) RHEL 7 server (target server). You will run the setup commands here.
-  - A backup from a 6.1, 6.2, or 6.3 Satellite server (source server) created with `katello-backup`. This backup can be with or without pulp-data, and can be from a RHEL 6 or 7 machine.
+  - A backup from a Satellite server (source server) created with `foreman-maintain` or `katello-backup`. This backup can be with or without pulp data, and can be from a RHEL 6 or 7 machine.
   - You will need a Satellite 6 subscription for the cloned machine. With the new Satellite Infrastructure [subscription model](https://access.redhat.com/solutions/3382781) you should have multiple Satellite subscriptions available.
 
 #### Setup ####
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -51,7 +51,7 @@
   when: enable_repos and disable_repos_result_fail
 
 - name: Enable required repos for Satellite installation
-  command: subscription-manager repos --enable rhel-{{ ansible_distribution_major_version }}-server-rpms --enable rhel-server-rhscl-{{ ansible_distribution_major_version }}-rpms --enable rhel-{{ ansible_distribution_major_version }}-server-satellite-{{ satellite_version }}-rpms
+  command: subscription-manager repos --enable rhel-{{ ansible_distribution_major_version }}-server-rpms --enable rhel-server-rhscl-{{ ansible_distribution_major_version }}-rpms --enable rhel-{{ ansible_distribution_major_version }}-server-satellite-{{ satellite_version }}-rpms --enable rhel-7-server-satellite-maintenance-6-rpms
   ignore_errors: True
   register: enable_repos_result
   when: enable_repos
@@ -60,7 +60,7 @@
   command: subscription-manager repos --enable rhel-7-server-satellite-{{ satellite_version }}-puppet{{ puppet_version }}-rpms
   when:
     - enable_repos
-    - puppet_version | int >= 4
+    - puppet_version|int == 4
 
 - name: set fact - enable_repos_result
   set_fact:
@@ -113,6 +113,12 @@
 - name: Install Satellite packages
   yum:
     name: "{{ satellite_package }}"
+    state: latest
+
+# due to support for remote databases in 6.4 the postgres user (and server) may be missing
+- name: Install postgresql-server
+  yum:
+    name: "postgresql-server"
     state: latest
 
 # The postgres user is created after installing postgresql packages, so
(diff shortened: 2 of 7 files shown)


jturel:
yee-haw !!

jturel:
yee-haw

README.md
```diff
@@ -1,15 +1,20 @@
...
 
 ## Getting Started
 Throughout this documentation, ensure that you understand the following terminology:
 - Source server: Existing Satellite server.
 - Target server: new server, to which Satellite server is being cloned.
 
+#### Supported Satellite Versions ####
+- 6.2
+- 6.3
+- 6.4
```
johnpmitsch: nice :+1:

roles/satellite-clone/tasks/main.yml
```diff
@@ -60,7 +60,7 @@
   command: subscription-manager repos --enable rhel-7-server-satellite-{{ satellite_version }}-puppet{{ puppet_version }}-rpms
   when:
     - enable_repos
-    - puppet_version | int >= 4
+    - puppet_version in [4]
```
johnpmitsch: Do we know what version of puppet sat 6.4 uses?

roles/satellite-clone/tasks/main.yml
```diff
@@ -115,6 +115,13 @@
     name: "{{ satellite_package }}"
     state: latest
 
+# Install postgres server
+# this was not installed by satellite RPM in 6.4
+- name: Install postgresql-server
+  yum:
+    name: "postgresql-server"
+    state: latest
```
johnpmitsch: Seems odd this wasn't installed by 6.4 rpm, did you ask the build team about it?

roles/satellite-clone/tasks/main.yml
```diff
@@ -115,6 +115,13 @@
     name: "{{ satellite_package }}"
     state: latest
 
+# Install postgres server
```
johnpmitsch: nitpick - First part of comment is redundant since its the name of the task (I know we do it some other places, but trying to break that pattern :) )

johnpmitsch:
I think we should add 6.4 to [the cleanup paused tasks task](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/tasks/main.yml#L235-L237). This was something that users asked for

The [--upgrade task is needed too](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/tasks/main.yml#L239-L243). The reason is we install from the CDN, which installs 6.4.latest, i.e. 6.4.1. If you have a backup that is from 6.4.0, you want to make sure the data is upgraded properly to run on 6.4.1.

roles/satellite-clone/tasks/main.yml
jturel: agreed - ansible is very self-documenting

roles/satellite-clone/tasks/main.yml
jturel: Talked w/ eric about it. He mentioned that the dependency was removed due to the support of remote dbs so installation of the postgresql server is handled by the installer. I decided to install the RPM here because postgres user was missing. This raises the question of what we will do for cloning a 6.4 sat with a remote DB. What do you think?

johnpmitsch: That makes sense, and ya, we do need the postgres user for certain tasks the way things currently are set up. 

I think a larger discussion about sat-clone and remote dbs is needed, but I imagine people would want support at some point. If we don't add it in this PR, we should update the docs to say cloning 6.4 installations with remote dbs isn't supported.

roles/satellite-clone/tasks/main.yml
jturel: Puppet 5 is what's on my box

johnpmitsch: So I'm guessing there is no puppet repo for 6.4?

jturel: Oh, right - it comes from the base sat repo now!

jturel:
yee-haw

jturel:
yee-haw

jturel:
@johnpmitsch got a green build - gonna run another now that it's squashed to make sure it stays that way :+1:

jturel:
yee-haw

jturel:
yee-haw

README.md
```diff
@@ -1,15 +1,20 @@
...
 
+#### Supported Satellite Versions ####
+- 6.2
+- 6.3
+- 6.4 (remote database cloning not currently supported)
+
 #### What you need: ####
   - A blank (vanilla install) RHEL 7 server (target server). You will run the setup commands here.
-  - A backup from a 6.1, 6.2, or 6.3 Satellite server (source server) created with `katello-backup`. This backup can be with or without pulp-data, and can be from a RHEL 6 or 7 machine.
+  - A backup from a Satellite server (source server) created with `katello-backup`. This backup can be with or without pulp data, and can be from a RHEL 6 or 7 machine.
```
johnpmitsch: one nitpicky thing is you could say `katello-backup` or `foreman-maintain` since the latter is new in 6.4

johnpmitsch: I guess we can keep the verbage around RHEL 6 or 7 here since we still support 6.2, but it is becoming less relevant. Do you think there is a better way to word it now?

roles/satellite-clone/tasks/main.yml
```diff
@@ -60,7 +60,7 @@
   command: subscription-manager repos --enable rhel-7-server-satellite-{{ satellite_version }}-puppet{{ puppet_version }}-rpms
   when:
     - enable_repos
-    - puppet_version | int >= 4
+    - puppet_version|int >= 4 and puppet_version|int < 5
```
johnpmitsch: Could you just say `puppet_version|int == 4`?

roles/satellite-clone/tasks/main.yml
```diff
@@ -170,7 +176,7 @@
   yum:
     name: puppetserver
     state: latest
-  when: puppet_version | int >= 4
+  when: puppet_version|int >= 4 and puppet_version|int < 5
```
johnpmitsch: same question here as above

johnpmitsch:
yee-haw

README.md
jturel: Good idea to mention foreman-maintain I think - will add that

roles/satellite-clone/tasks/main.yml
jturel: Looking at how puppet_version is derived ... I think it will surely work!

jturel:
yee-haw

johnpmitsch:
Thanks @jturel!

roles/satellite-clone/tasks/main.yml
```diff
@@ -115,6 +115,12 @@
     name: "{{ satellite_package }}"
     state: latest
 
+# due to support for remote databases in 6.4 the postgres user (and server) may be missing
```
sthirugn: Mongodb can also be external right, any reason why it is not checked explicitly here like postgres?

jturel: That hadn't occurred to me at the time :(

PR #350, opened 2019-01-09
Reset puppet server ssl chain on install

### LINKED ISSUE #349: 6.4 issue with puppet ssl chain
Details in https://bugzilla.redhat.com/show_bug.cgi?id=1664350

This seems to be on 6.3 to 6.4 upgrades

johnpmitsch:
Fixes #349

### DIFF
--- a/roles/satellite-clone/vars/satellite_6.4.yml
+++ b/roles/satellite-clone/vars/satellite_6.4.yml
@@ -3,6 +3,6 @@ satellite_package: satellite
 satellite_installer_cmd: satellite-installer
 satellite_scenario: satellite
 capsule_puppet_module: foreman-proxy
-satellite_installer_options: "--foreman-ipa-authentication false --disable-system-checks"
+satellite_installer_options: "--foreman-ipa-authentication false --reset-puppet-server-ssl-chain-filepath --disable-system-checks"
 satellite_upgrade_options: "--disable-system-checks"
 verify_rake_task: reimport


johnpmitsch:
yee-haw

beav:
how is the SSL chain filepath getting altered to begin with? The fix looks correct but I'm not sure what the root cause is

ekohl:
@beav this path is stored in the answers file and is correct for Puppet 3. When moving to Puppet 4 we have a hook that resets a bunch of answers (https://github.com/Katello/katello-installer/blob/KATELLO-3.4/hooks/pre_validations/31-upgrade-puppet.rb). It doesn't reset this one. The hook was based on https://projects.theforeman.org/projects/foreman/wiki/Upgrading_from_Puppet_3_to_4 but that doesn't mention it.

On upgrading systems this isn't really a problem because the hook copies the certs (https://github.com/Katello/katello-installer/blob/918d3780558b4769da0a32b965cbc5369c2c544a/hooks/init/31-upgrade-puppet.rb#L29) so the old location remains valid. Realistically: how often do you have a new SSL chain on your Puppet master? Probably once every 5 years when your CA expires (maybe this was extended in later versions).

When cloning the old location isn't available. I assume the clone does present ensure the certificates are present in the new (default) location. By resetting the answer the puppetserver knows where to find the file.

We should really fix it properly for all systems because when you do want to replace the CA, you can see the same breakage. That's why I consider this a workaround

johnpmitsch:
@beav expanding on what @ekohl said, It does seem some further fixes are needed, but we haven't seen this issue (afaik) on upgraded puppet systems. The workflow to create this error was: clone 6.3 backup; upgrade puppet 3 to 4; upgrade to sat 6.4; run a backup; and clone that backup. In this case, you have two backups being restored, so I'm guessing that caused (or exposed) the mismatch in the puppet answers file.

After we hear if this fixed the automation issue, we can file a proper installer bug for the more permanent fix.

johnpmitsch:
yee-haw

ntkathole:
@johnpmitsch yup, I tested it with above changes and it works :100:  Thanks :+1:

johnpmitsch:
Thanks for testing @ntkathole! 

1/6 dolly clones failed, looks like it was a resources issue again, but it worked when I re-ran it. It was a 6.3 clone too so unlikely affected by this PR. merging!

Issue #351, opened 2019-02-27
Add support for cloning satellite 6.5


pondrejk:

johnpmitsch:
this has been merged!

PR #352, opened 2019-02-27
cloning sat 6.5

pondrejk:
- related to #351 
- satellite_6.5.yml is just a copy of the 6.4 version
- I was able to do a successful clone using this patch


johnpmitsch:
@pondrejk I updated jenkins with online, offline, and offline-skip-pulp backups for 6.5 and it worked well! The supported Satellite versions in the README has to be updated too, but we can do that in a follow-up. Thanks for the PR!

PR #356, opened 2019-05-07
make the satellite-clone script python3 compatible

evgeni:


johnpmitsch:
@evgeni are we always shipping python3 with satellite-clone? i.e. if I install sat 6.3 and install sat-clone from satellite-maintenance repo, will I be using python3?

johnpmitsch:
also, does this mean the script is both python2 + 3 compatible? 

Some background on this change might clear up my questions, I've been out of the loop for a bit so may have missed some discussions

evgeni:
No, today it is using py2. However, on rhel8, we will have to use py3. So this change is for the future us.

And yes, this make the script work with both pythons.

PR #375, opened 2020-10-09
Remote database support

johnpmitsch:


docs/satellite-clone.md
```diff
@@ -91,3 +91,25 @@ On the target server:
       # ansible-playbook satellite-clone-playbook.yml
     ```
   The playbook run may may take a while to complete.
+
+## Restoring a Satellite With a Remote Database
+
+Note: the backup used for this procedures must have the correct options to manage remote databases in `/etc/foreman-installer/scenarios.d/satellite-answers.yaml`. Please check that the answers file has the correct database management values filled.
```
johnpmitsch, line 97: probably should actually check this in the playbook, the `manage-db` type options not matching in the backup and the target satellite will cause issues. For instance, if I have a backup that doesn't have foreman-manage-db set to true, but I pass in that option to clone, it will install foreman remotely, but then `foreman-maintain` will restore the backup, see that foreman-manage-db is false, and try to set up the database locally. This is because foreman-maintain uses `satellite-installer --reset`

johnpmitsch:
it worked! :tada:

```
TASK [satellite-clone : debug] *******************************************************************************************************************************************************************
Friday 09 October 2020  22:54:21 +0000 (0:00:00.390)       0:49:22.683 ******** 
ok: [localhost] => {
    "msg": "****NOTE**** Your Satellite's hostname is updated to match the original Satellite ****NOTE**** Your Satellite's password is updated to changeme"
}

msg: ****NOTE**** Your Satellite's hostname is updated to match the original Satellite ****NOTE**** Your Satellite's password is updated to changeme

PLAY RECAP ***************************************************************************************************************************************************************************************
localhost                  : ok=74   changed=25   unreachable=0    failed=0   
```

I still would like to clean up a couple things and have an open question or two, but moving to ready for review as it is testable.

roles/satellite-clone/tasks/main.yml
```diff
@@ -250,16 +259,8 @@
...
-    command: sed -n -e 's/^.*myDS.password=//p' /etc/candlepin/candlepin.conf
-    register: candlepin_password
-
-  - set_fact:
-      clone_candlepin_password: "{{ candlepin_password.stdout }}"
-      cacheable: true
-
   - name: Migrate candlepin db
-    command: "/usr/share/candlepin/cpdb --update -p {{ clone_candlepin_password }}"
+    command: "/usr/share/candlepin/cpdb --update --dbhost={{ candlepin_db_host.value }} --password={{ candlepin_db_password.value }}"
```
johnpmitsch, line 263: Not sure if we should pass more options here. On one hand, we've only passed the password for a while and haven't seen any issues, on the other hand foreman-installer passes more options https://github.com/theforeman/puppet-candlepin/blob/f32c2e7d9c7326a24a5795714a57bbc18c1e8e33/lib/puppet/provider/cpdb_update/cpdb_update.rb#L21-L28

This cpdb step was originally done to makes sure the backup is brought up to the latest z-version of satellite that is installed, for instance if the backup is 6.6.1 and the target server's sat version is 6.6.3, we want any changes in the candlepin db to be applied. This is likely being done in satellite-installer --upgrade step now, which we have for the same reason. 

I'm not sure if there was a reason the cpdb step was added before that step though, so would like to check a few things and confirm my assumptions before removing or gating to older versions.

roles/satellite-clone/tasks/main.yml
```diff
@@ -250,16 +259,8 @@
   - name: Wait for foreman-tasks service to start
     command: sleep 300
 
-  - name: Get candlepin credentials
-    command: sed -n -e 's/^.*myDS.password=//p' /etc/candlepin/candlepin.conf
-    register: candlepin_password
-
-  - set_fact:
-      clone_candlepin_password: "{{ candlepin_password.stdout }}"
-      cacheable: true
```
johnpmitsch, line 259: This 'caching facts' thing was to try to make it so sat-clone could start at a step after a failed or incomplete run and still have all the context, but it really doesn't work well in practice and I haven't seen anyone use it. I don't see any value in keeping it, ideally we remove the whole feature.

docs/satellite-clone.md
```diff
@@ -91,3 +91,24 @@ On the target server:
       # ansible-playbook satellite-clone-playbook.yml
     ```
   The playbook run may may take a while to complete.
+
+## Cloning a Satellite With Remote Databases
+
+Note: the backup used for this procedures must have the correct options to manage remote databases in `/etc/foreman-installer/scenarios.d/satellite-answers.yaml`. Please check that the answers file has the correct database management values filled configuration backups.
```
sthirugn, line 97: Can you elaborate on what the user actually needs to check here?

sthirugn, line 97: s/the/The

docs/satellite-clone.md
```diff
@@ -91,3 +91,24 @@ On the target server:
...
+
+1. Set up the remote databases according to Satellite documentation in the same way that they were set up on your original Satellite.
```
sthirugn, line 99: perhaps hyperlink to https://access.redhat.com/documentation/en-us/red_hat_satellite/ ?

sthirugn:
Great work. @johnpmitsch

docs/satellite-clone.md
johnpmitsch: ya, I can specify the specific ones. Unfortunately the confusion is the result of a docs bug: https://bugzilla.redhat.com/show_bug.cgi?id=188784

johnpmitsch: Updated, I mentioned the BZ and the specific options.

johnpmitsch:
I'm wondering if we should have some sort of warning about restoring a clone with remote databases. We do recommend all clones happen on an isolated network, but if they aren't able to isolate the network, the clone could try to drop the production database as the hostname will be the same. 

I'm thinking we should just caveat the whole procedure as "experimental" and "at your own risk" while mentioning the risk of the clone talking to the production databases. This procedure likely will stay only in the upstream documentation until it's a more polished process.

@sthirugn @jturel thoughts?

jturel:
@johnpmitsch good catch. how about a prompt which alludes to the severity of the implications of not being on an isolated network?

johnpmitsch:
> @johnpmitsch good catch. how about a prompt which alludes to the severity of the implications of not being on an isolated network?

Are you thinking for all clones or the ones who are cloning with remote databases? The latter may be harder to do as it's all based on passing additional installer options in, which wouldn't necessarily mean a remote database clone.

docs/satellite-clone.md
```diff
@@ -91,3 +91,24 @@ On the target server:
       # ansible-playbook satellite-clone-playbook.yml
     ```
   The playbook run may may take a while to complete.
+
+## Cloning a Satellite With Remote Databases
+
+Note: The backup used for these procedures must have the correct options to manage remote databases in `/etc/foreman-installer/scenarios.d/satellite-answers.yaml`. Please check that the answers file has the correct database management values filled in the configuration backup file. These will be values such as `db-manage` in the `foreman` section and `pulp-manage-db` in the `katello` section. They are currently (at the time of this writing) [missing from the initial setup documentation](https://bugzilla.redhat.com/show_bug.cgi?id=1887846)
```
jturel, line 97: I think this puts a large burden on the user to work around our own bug. Couldn't we handle this during the playbook run?

docs/satellite-clone.md
```diff
@@ -91,3 +91,24 @@ On the target server:
...
+
+1. Set up the remote databases according to [Satellite documentation](https://access.redhat.com/documentation/en-us/red_hat_satellite/) in the same way that they were set up on your original Satellite.
+2. Set up a base RHEL7 server that will become the cloned satellite and make sure the target clone server has network access to the remote databases.
```
jturel, line 100: Step 2 and 1 should swap places because current 1 implies that would-be satellite server already exists

docs/satellite-clone.md
```diff
@@ -91,3 +91,24 @@ On the target server:
...
+3. The remote databases for the clone will also need to be reached on the same hostname as ones used for the original Satellite. You can use `/etc/hosts` on the target server to associate the original remote database hostnames with the new clone database IP addresses.
+4. Install satellite-clone on target server according to [instructions](#instructions), but don't run the playbook yet.
+5. In `satellite-clone-vars.yml` find and uncomment the line: `installer_additional_options: ""`
```
jturel, line 103: Don't the installer opts come from last_scenario.yaml? IOW why would the user need to set these? If they must be set, I think it'd be nice to create the string during the playbook run, and let the user specify key-value pairs - just thinking about the UX here.

roles/satellite-clone/tasks/main.yml
```diff
@@ -19,6 +19,20 @@
     keys: [":foreman_url"]
   register: backup_hostname
 
+- name: Get candlepin password
```
jturel, line 22: I'm not sure how many times we do this but it seemed a bit slow (I hope I'm not confusing it w/ another step). Maybe we can just load the entirety of any yamls we need once, and refer to any keys later directly in memory?

roles/satellite-clone/tasks/main.yml
jturel: +1 to removing or gating as appropriate. It certainly gives the appearance of not being necessary.

jturel (commented):
The feedback I left is mostly to the effect of how I'd like the user experience to be:

1. set up remote servers per Red Hat documentation
2. run the playbook and it just works with minimal extra configuration or steps

We see that it works, so why not try to optimize the manual aspects away where feasible?

jturel:
> > @johnpmitsch good catch. how about a prompt which alludes to the severity of the implications of not being on an isolated network?
> 
> Are you thinking for all clones or the ones who are cloning with remote databases? The latter may be harder to do as it's all based on passing additional installer options in, which wouldn't necessarily mean a remote database clone.

KISS and do it all the time would be my approach!

docs/satellite-clone.md
johnpmitsch: I think the big part that is missing is to not install Satellite, just set up the remote dbs and a blank RHEL7 server. I'll find a better way to word this!

roles/satellite-clone/tasks/main.yml
johnpmitsch: Do you have any numbers on how long it took? It was around 1 second for me but it was a really small backup, I am curious how it's performing on a large backup. I'm sure we could improve it but will have to see what it takes.

docs/satellite-clone.md
johnpmitsch: Good question, I only tested out running things on a single server (no remote dbs) with the current sat-clone, but I didn't try it with remote db's set up properly. We do restore the config before the first install. 

I'll test it out now that we know how to set up the db's properly!

roles/satellite-clone/tasks/main.yml
johnpmitsch: I'm not too clear on which versions of the installer are doing what, here are two commits in that area:
https://github.com/theforeman/foreman-installer/commit/286456500c911f3aa997b8b16cb597e67cbd2ba6#diff-b8060b8893bfb1a23cadc8c756
https://github.com/theforeman/puppet-candlepin/commit/fadb503a9c7fc12ac221f1dd74c0792540c33750

I can look into it more, but will keep this step for now.

jturel (approved):
ACK!

Process is well documented and I tested a backup from a large environment with remote databases. It all checked out. Awesome!

sthirugn:
@johnpmitsch great work getting this in :+1:

PR #377, opened 2020-11-10
Fixes #376 - Disassociate capsules from lifecycle environments

### LINKED ISSUE #376: Disassociate capsules from lifecycle environments fails with db error
satellite-clone fails at disassociate capsule step with below error:
~~~
2020-11-10 11:02:49,593 p=48208 u=root |  fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["psql", "foreman", "-c", "delete from katello_capsule_lifecycle_environments where capsule_id in (select smart_proxy_id from features_smart_proxies where feature_id = (select id from features where features.name = 'Pulp Node'));"], "delta": "0:00:00.012455", "end": "2020-11-10 11:02:49.575174", "msg": "non-zero return code", "rc": 1, "start": "2020-11-10 11:02:49.562719", "stderr": "ERROR:  relation \"features_smart_proxies\" does not exist\nLINE 1: ...s where capsule_id in (select smart_proxy_id from features_s...\n                                                             ^", "stderr_lines": ["ERROR:  relation \"features_smart_proxies\" does not exist", "LINE 1: ...s where capsule_id in (select smart_proxy_id from features_s...", "     
~~~

anandkumaragrawal:


roles/satellite-clone/tasks/main.yml
```diff
@@ -348,7 +348,7 @@
   # This is to prevent the clone trying to sync with existing original capsules. We do this with SQL in case
   # hammer is not able to execute a command.
   - name: Disassociate capsules from lifecycle environments to avoid communication with production capsules
-    command: psql foreman -c "delete from katello_capsule_lifecycle_environments where capsule_id in (select smart_proxy_id from features_smart_proxies where feature_id = (select id from features where features.name = 'Pulp Node'));"
+    command: psql foreman -c "delete from katello_capsule_lifecycle_environments where capsule_id in (select smart_proxy_id from smart_proxy_features where feature_id = (select id from features where features.name = 'Pulp Node'));"
```
johnpmitsch, line 351: What Satellite version did this table name change in? 

Sat-clone has to stay backwards compatible so likely will have to set this dynamically based on sat version

anandkumaragrawal: @johnpmitsch When I test on Satellite version 6.5, 6.6, 6.7, 6.8 all have table name smart_proxy_features. I am not really sure if the name was `features_smart_proxies` in any version

johnpmitsch: > @johnpmitsch When I test on Satellite version 6.5, 6.6, 6.7, 6.8 all have table name smart_proxy_features. I am not really sure if the name was `features_smart_proxies` in any version

It looks like the table name changed here https://github.com/theforeman/foreman/blob/9080d4e5bc4fe7c7628cc1fce738302aceaed7ce/db/migrate/20180705230311_smart_proxy_capabilities.rb#L3

That was introduced in Foreman 1.22 which I believe is 6.5, if we can confirm that we can update to dynamically set it to use the new table name in 6.5 and later

anandkumaragrawal: @johnpmitsch Thank you. You are right, I have checked on Satellite 6.5 again, and I can see the table name was `features_smart_proxies` which changed after that.

evgeni: 6.5 is 1.20, 6.6 is 1.22.

johnpmitsch: @evgeni thanks for confirming, sounds like we landed on:
 - 6.5 *and* earlier should use `features_smart_proxies`
 - 6.6 *and* later should use `smart_proxy_features`

@anandkumaragrawal I would recommend setting the table name as a variable like `sp_features_table_name`. Make the default `smart_proxy_features` [here](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/defaults/main.yml) and set the templates 6.5 and earlier [here](https://github.com/RedHatSatellite/satellite-clone/tree/master/roles/satellite-clone/vars) to `features_smart_proxies`. Then you can use the variable in this command

jturel:
Looks like this needs a rebase now :)

anandkumaragrawal:
This is already fixed with PR#380, Hence closing this.

PR #382, opened 2021-01-08
On 6.8+ install selinux packages

chris1984:
They were dropped as dependicnes of the satellite package which now causes the untar operation to fail.

Worked fine on my test box, will kick off dolly on the pr and post the results.

### DIFF
--- a/roles/satellite-clone/defaults/main.yml
+++ b/roles/satellite-clone/defaults/main.yml
@@ -17,3 +17,4 @@ overwrite_etc_hosts: True
 puppet_version: 3
 check_networking_interfaces: True
 sp_features_table_name: smart_proxy_features
+selinux_packages: []
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -184,6 +184,13 @@
     state: latest
   when: puppet_version|int == 4
 
+  # Because of https://projects.theforeman.org/issues/30506
+- name: Install Satellite SELinux packages for 6.8 and above
+  yum:
+    name: "{{ selinux_packages | join(',') }}"
+    state: present
+  when: selinux_packages | length > 0
+
 - name: untar config files (for cloning only)
   command: tar --selinux --overwrite -xf {{ backup_dir }}/config_files.tar.gz -C / --exclude=var/lib/foreman/public
 
--- a/roles/satellite-clone/vars/satellite_6.8.yml
+++ b/roles/satellite-clone/vars/satellite_6.8.yml
@@ -6,3 +6,7 @@ capsule_puppet_module: foreman-proxy
 satellite_installer_options: "--foreman-ipa-authentication false --reset-puppet-server-ssl-chain-filepath --disable-system-checks"
 satellite_upgrade_options: "--disable-system-checks"
 verify_rake_task: reimport
+selinux_packages:
+  - crane-selinux
+  - foreman-selinux
+  - katello-selinux


roles/satellite-clone/tasks/main.yml
```diff
@@ -184,8 +184,19 @@
...
+    name:
+      - crane-selinux
+      - foreman-selinux
+      - katello-selinux
+    state: present
+  when: satellite_version in ["6.8"]
+
 - name: untar config files (for cloning only)
   command: tar --selinux --overwrite -xf {{ backup_dir }}/config_files.tar.gz -C / --exclude=var/lib/foreman/public
+  when: satellite_version in ["6.8"]
```
sthirugn, line 199: why is this restricted to `6.8` only?

sthirugn (changes requested):

chris1984:
Because on 6.7 and below the package is a dep of satellite and would
already be installed.

roles/satellite-clone/tasks/main.yml
chris1984: Because on 6.7 and below the package is a dep of satellite and would
already be installed.

sthirugn: I understand about the new task `name: Install Satellite SELinux packages for 6.8 and above`

But this step below, if I understand correctly is general for all versions. If you restrict it to 6.8 only, older versions won't untar isn't it?
```
- name: untar config files (for cloning only)
  command: tar --selinux --overwrite -xf {{ backup_dir }}/config_files.tar.gz -C / --exclude=var/lib/foreman/public`
```

chris1984: Oh my bad, I must have failed during copy/paste. me-- I will fix that and push up my changes and do another dolly run.

chris1984: Nice catch

chris1984: @sthirugn Updated. Starting dolly run

roles/satellite-clone/tasks/main.yml
```diff
@@ -184,6 +184,16 @@
     state: latest
   when: puppet_version|int == 4
 
+  # Because of https://projects.theforeman.org/issues/30506
+- name: Install Satellite SELinux packages for 6.8 and above
```
jturel, line 188: There's a lot of checks in this module relating to the `satellite_version`. A more flexible and generic way would be to define an array in the respective vars files called something like `selinux_packages` which would be empty for versions other than 6.8. That means this task would be more along the lines of "install any specific selinux packages" and we don't care about satellite version at all. That also gives us more fine-grained control per satellite version!

Ultimately I'd love to see us move in that direction so that the playbooks become quite a bit 'dumber'

jturel: The pattern I suggest is also an improvement because when it's time to support 6.9 and we copy the 6.8 vars yaml we can review it (perhaps adding more -selinux packages for example) to ensure we don't miss anything. The way this is currently written would require review of this playbook and hopefully not missing this

roles/satellite-clone/tasks/main.yml
sthirugn: thanks

roles/satellite-clone/tasks/main.yml
sthirugn: I like the idea

jturel: Thought about it some more. In this particular case maybe it's fair enough to just have it ensure these packages are installed for any version. Downside there would be if the list should actually change between versions in the future. then we'd need something like I suggested above

chris1984:
@jturel @sthirugn updated and removed the 6.8 version condition on the SELinux package install based on your latest comment. Let me know if that looks good or any other changes need to be made. If it looks good then I will kick off the dolly tests on the pr

jturel:
LGTM! Let's see if dolly is happy with it

roles/satellite-clone/tasks/main.yml
sthirugn: hey @jturel do you mind creating a github issue to think about it in future, I do like the idea of keeping the version-specific changes under control and separate.

sthirugn (approved):

roles/satellite-clone/tasks/main.yml
```diff
@@ -184,6 +184,15 @@
...
```
johnpmitsch, line 188: Is this actually meant to be for 6.8 and above? This would run on all versions. 

If the packages are expected to be present for all versions that is ok, just want to double check that's the case :monocle_face:

roles/satellite-clone/tasks/main.yml
jturel: Done! https://github.com/RedHatSatellite/satellite-clone/issues/383

roles/satellite-clone/tasks/main.yml
jturel: Technically speaking this is for 6.8 and above but the idea is that there shouldn't be harm in ensuring the packages are present when cloning 6.7 and below (they should already be there due to RPM deps for older versions).

johnpmitsch: sounds good :+1:

johnpmitsch (approved):

roles/satellite-clone/tasks/main.yml
chris1984: It looks like in 6.7 we do not have a crane package, so the job is failing with:

```bash
2021-01-11 20:07:02,915 p=15208 u=root n=ansible | fatal: [localhost]: FAILED! => {"changed": false, "msg": "No package matching 'crane-selinux' found available, installed or updated", "rc": 126, "results": ["No package matching 'crane-selinux' found available, installed or updated"]}
```

Looking at a 6.7 box all I see for crane is:

```bash
[root@satellite1 ~]# rpm -qa | grep crane
python2-crane-3.3.1-9.el7sat.noarch
```

Should we change this to just run on 6.8 or go with the var defined in the yaml file option?

jturel: There are two options imo:

1. Write this command such that it won't fail the ansible run. Probably not the best solution but I suppose it would work
2. Yep. Do what I outlined here https://github.com/RedHatSatellite/satellite-clone/issues/383

johnpmitsch: > 2\. Yep. Do what I outlined here #383

+1, you can make the default a blank array [here](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/defaults/main.yml) and set it to the expected packages [here](https://github.com/RedHatSatellite/satellite-clone/blob/master/roles/satellite-clone/vars/satellite_6.8.yml)

It looks like you can use a comma separated list of packages with the [yum module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html)

>To operate on several packages this can accept a comma separated string of packages or (as of 2.0) a list of packages.

So you may just be able to pass the variable and it will work fine, or can join with a filter e.g. `{{ selinux_pkgs | join(",") }}`

You will probably need to skip the step for a blank array, so you could do something like `when: selinux_pkgs | length > 0` iirc

chris1984:
@johnpmitsch how does this look? If it looks fine I will kick off a test with Dolly.

roles/satellite-clone/defaults/main.yml
```diff
@@ -17,3 +17,4 @@ overwrite_etc_hosts: True
 puppet_version: 3
 check_networking_interfaces: True
 sp_features_table_name: smart_proxy_features
+selinux_pkgs:
```
johnpmitsch, line 20: @chris1984 this just needs to be a blank array, also `selinux_packages` would be a better name imo (I know I shortened it in the example)

roles/satellite-clone/tasks/main.yml
```diff
@@ -184,6 +184,13 @@
...
+  yum:
+    name: "{{ selinux_pkgs | join(','') }}"
```
johnpmitsch, line 190: Looks like there is an extra quote here `join(','')`, should be `join(',')`

chris1984: Damn VScode trying to be smart lol

roles/satellite-clone/defaults/main.yml
chris1984: Fixed

roles/satellite-clone/tasks/main.yml
chris1984: Fixed

johnpmitsch:
@chris1984 looks good, go4it!

chris1984:
@johnpmitsch results from Dolly:

```
PLAY RECAP *********************************************************************
192.168.121.109            : ok=21   changed=16   unreachable=0    failed=0    skipped=3    rescued=0    ignored=0
192.168.121.212            : ok=21   changed=16   unreachable=0    failed=0    skipped=3    rescued=0    ignored=0
192.168.121.244            : ok=21   changed=16   unreachable=0    failed=0    skipped=3    rescued=0    ignored=0
192.168.121.246            : ok=21   changed=16   unreachable=0    failed=0    skipped=3    rescued=0    ignored=0
```

jturel (approved):
Untested but ACKing based on dolly results

johnpmitsch:
@chris1984 thanks!

Issue #268, opened 2017-11-20
disable IPA authentication


johnpmitsch:
run the installer with --foreman-ipa-authentication false

see here for context: 
https://bugzilla.redhat.com/show_bug.cgi?id=1513949

ekohl:
Looks like this fix was incomplete. The Smart Proxy can also have a realm provider which may cause the Smart Proxy to fail.

PR #386, opened 2021-06-08
Add support for Satellite 6.9

jturel:


roles/satellite-clone/vars/satellite_6.9.yml
```diff
@@ -0,0 +1,16 @@
+---
+satellite_package: satellite
+satellite_installer_cmd: satellite-installer
+satellite_scenario: satellite
+capsule_puppet_module: foreman-proxy
+satellite_installer_options: "--foreman-ipa-authentication false --reset-puppet-server-ssl-chain-filepath --disable-system-checks"
+satellite_upgrade_options: "--disable-system-checks"
+verify_rake_task: reimport
+selinux_packages:
```
ekohl, line 9: We also have pulpcore-selinux now. Should that be listed as well?

roles/satellite-clone/vars/satellite_6.9.yml
```diff
@@ -0,0 +1,16 @@
+---
+satellite_package: satellite
+satellite_installer_cmd: satellite-installer
+satellite_scenario: satellite
+capsule_puppet_module: foreman-proxy
+satellite_installer_options: "--foreman-ipa-authentication false --reset-puppet-server-ssl-chain-filepath --disable-system-checks"
```
ekohl, line 6: Why are these 2 options always passed in? (I understand disabling system checks, at least party)

roles/satellite-clone/vars/satellite_6.9.yml
jturel: I think it makes sense to add it. We needed to ensure these were installed beforehand because of https://projects.theforeman.org/issues/30506

roles/satellite-clone/vars/satellite_6.9.yml
jturel: foreman-ipa-authentication false -> #268 
reset-puppet-server-ssl-chain-filepath -> #349

roles/satellite-clone/vars/satellite_6.9.yml
jturel: I see we didn't add candlepin-selinux either. perhaps let's just see if it works as-is?

ekohl: I don't know satellite-clone that well. If it runs the installer, that should generally take care of installing the packages, right?

roles/satellite-clone/vars/satellite_6.9.yml
ekohl: > foreman-ipa-authentication false -> #268

I guess that makes sense.

>  reset-puppet-server-ssl-chain-filepath -> #349

Hah, just last Friday I was playing with this: https://github.com/theforeman/puppet-puppet/pull/779/commits/c05e0a9c8111a7cb27c2f3f25675e5584ff0c261

That said, it very much feels like a 6.3 -> 6.4 change due to Puppet 3 -> 4. In current versions I don't think it would be needed.

jturel: > Hah, just last Friday I was playing with this: [theforeman/puppet-puppet@c05e0a9](https://github.com/theforeman/puppet-puppet/commit/c05e0a9c8111a7cb27c2f3f25675e5584ff0c261)
> 
Beautiful coincidence!

> That said, it very much feels like a 6.3 -> 6.4 change due to Puppet 3 -> 4. In current versions I don't think it would be needed.

Let's give it a whirl (about to start a test of this PR right now to see)

wbclark: > foreman-ipa-authentication false -> #268
> reset-puppet-server-ssl-chain-filepath -> #349

TYVM. Could you add this as a comment in the YAML so that future us / future somebody else can find it more easily?

roles/satellite-clone/vars/satellite_6.9.yml
ekohl: That was also something I suspected.

roles/satellite-clone/vars/satellite_6.9.yml
jturel: Can do. I feel a utility such as this (expected to run on old, old versions of Satellite) warrants excessive commenting

jturel:
New one for me here:

```
--> Finished Dependency Resolution                                                                                                                            
Error: Package: satellite-common-6.9.2-1.el7sat.noarch (rhel-7-server-satellite-6.9-rpms)                                                                     
           Requires: ansible >= 2.9                                                                                                                           
           Installed: ansible-2.4.2.0-2.el7.noarch (@rhel-7-server-extras-rpms)                                                                               
               ansible = 2.4.2.0-2.el7                                                                                                                        
           Available: ansible-2.4.0.0-1.el7ae.noarch (rhel-7-server-satellite-maintenance-6-rpms)                                                             
               ansible = 2.4.0.0-1.el7ae        
```

Which repo should ansible > 2.9 be coming from? Need to update the list to auto-enable it

sthirugn:
Looks good to me, but I don't have time to test it. Feel free to merge after @ekohl acks it.

evgeni:
> New one for me here:
> 
> ```
> --> Finished Dependency Resolution                                                                                                                            
> Error: Package: satellite-common-6.9.2-1.el7sat.noarch (rhel-7-server-satellite-6.9-rpms)                                                                     
>            Requires: ansible >= 2.9                                                                                                                           
>            Installed: ansible-2.4.2.0-2.el7.noarch (@rhel-7-server-extras-rpms)                                                                               
>                ansible = 2.4.2.0-2.el7                                                                                                                        
>            Available: ansible-2.4.0.0-1.el7ae.noarch (rhel-7-server-satellite-maintenance-6-rpms)                                                             
>                ansible = 2.4.0.0-1.el7ae        
> ```
> 
> Which repo should ansible > 2.9 be coming from? Need to update the list to auto-enable it
> 
> Is it `rhel-7-server-ansible-2.9-rpms` ?

Exactly this.

roles/satellite-clone/tasks/main.yml
```diff
@@ -60,6 +60,10 @@
   register: enable_repos_result
   when: enable_repos
 
+- name: Enable Ansible 2.9 repository
+  command: subscription-manager repos --enable rhel-{{ ansible_distribution_major_version}}-server-ansible-2.9-rpms
```
evgeni, line 64: The el8 repo has a completely different name, so no need to template it here

roles/satellite-clone/tasks/main.yml
```diff
@@ -60,6 +60,10 @@
...
+  when: satellite_version in ["6.9"]
```
evgeni, line 65: This can (and should) be done on 6.8 too, we just didn't have the more explicit dependency in the RPM that would trigger the error you posted in 6.8

roles/satellite-clone/tasks/main.yml
```diff
@@ -197,7 +201,7 @@
   file:
     path: /var/opt/rh/rh-postgresql12/lib/pgsql/data/postgresql.conf
     state: absent
-  when: satellite_version in ["6.8"]
+  when: satellite_version in ["6.8", "6.9"]
```
evgeni, line 204: I wonder of we should check `satellite_version is version("6.8", ">=")` instead, to allow for 6.10 and whatever.

See https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#comparing-versions

roles/satellite-clone/tasks/main.yml
jturel: I was testing this as-is and I ran into an error in the middle of execution and I think it had to with the fact that I started the playbook while running ansible 2.4 and ansible was upgraded from underneath: 

```
TASK [satellite-clone : Install packages necessary to check DB status] ***************************************************************************************
Tuesday 08 June 2021  15:28:19 +0000 (0:07:30.279)       0:10:16.888 **********  
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'Task' object has no attribute 'async_val'
fatal: [localhost]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}

msg: Unexpected failure during module execution.

```
This step completed successfully when I ran it a second time (with ansible 2.9 loaded from the start)

To me it seems that the initial steps in the doc should reflect the correct repo to enable to get the right ansible version installed before the clone process is started

roles/satellite-clone/tasks/main.yml
jturel: That's nice. wonder why it wasn't used in several places in this repo...

evgeni: If I'd tell you, I'd owe you drinks, so you can forget again ;)

roles/satellite-clone/tasks/main.yml
evgeni: I don't understand where this issue comes from, but if using 2.9 fixes it, we should be good.

evgeni: Interestingly, docs don't say to enable Ansible repos: https://access.redhat.com/documentation/en-us/red_hat_satellite/6.9/html/upgrading_and_updating_red_hat_satellite/cloning_satellite_server#sec-Cloning_to_Target

roles/satellite-clone/tasks/main.yml
jturel: `version` was changed from `version_compare` in 2.5 so it seems like we can't use this option since users might be on 2.4 for older sats

evgeni: Satellite 6.5 used Ansible 2.6: https://access.redhat.com/documentation/en-us/red_hat_satellite/6.5/html/installing_satellite_server_from_a_connected_network/installing_satellite_server#configuring_repositories_satellite

We don't support anything older I believe.

evgeni: I think the main question is: is this supposed to work without the "correct" Ansible repo enabled? Keeping 2.4 compatibility seems insane to me.

CC @mccun934

jturel: I've been following the README in this repo :-/

The docs you linked are incorrect for not specifying which ansible repo to enable, right?

evgeni: I honestly don't know. For normal Sat operation, you need the *matching* Ansible repo. But s-clone is special.

It does enable the "right" repos itself:
https://github.com/RedHatSatellite/satellite-clone/blob/f374e242911c5c41e9f6e65cb173530f0f948356/roles/satellite-clone/tasks/main.yml#L58

But actually it doesn't, as it doesn't (before this PR) as it didn't enable Ansible at all...

roles/satellite-clone/tasks/main.yml
evgeni: But you could template the Ansible version, and put that into vars :)

README.md
```diff
@@ -27,10 +28,17 @@ On the target server:
...
+   ```console
+     # subscription-manager repos --enable rhel-7-server-ansible-2.9-rpms
+     # yum install ansible
+   ```
+
+   Otherwise:
    ```console
      # subscription-manager repos --enable rhel-7-server-extras-rpms
-     # yum install -y ansible
+     # yum install ansible
```
evgeni, line 41: Nobody should install ansible from Extras these days

jturel: I'm going to update this with some sort of matrix (satversion, ansible version)

jturel: @evgeni have a look now!

roles/satellite-clone/tasks/main.yml
jturel: Since I haven't added any _new_ checks of this nature I'd like to keep the current approach

roles/satellite-clone/tasks/main.yml
```diff
@@ -55,7 +55,7 @@
   when: enable_repos and disable_repos_result_fail
 
 - name: Enable required repos for Satellite installation
-  command: subscription-manager repos --enable rhel-{{ ansible_distribution_major_version }}-server-rpms --enable rhel-server-rhscl-{{ ansible_distribution_major_version }}-rpms --enable rhel-{{ ansible_distribution_major_version }}-server-satellite-{{ satellite_version }}-rpms --enable rhel-7-server-satellite-maintenance-6-rpms
+  command: subscription-manager repos --enable rhel-{{ ansible_distribution_major_version }}-server-rpms --enable rhel-server-rhscl-{{ ansible_distribution_major_version }}-rpms --enable rhel-{{ ansible_distribution_major_version }}-server-satellite-{{ satellite_version }}-rpms --enable rhel-7-server-satellite-maintenance-6-rpms --enable {{ ansible_repository_label }}
```
evgeni, line 58: why the indirection via `ansible_repository_label`, instead of using `rhel-7-server-ansible-{{ satellite_ansible_version }}-rpms` directly?

more curious than bothered :)

roles/satellite-clone/vars/satellite_6.8.yml
```diff
@@ -14,3 +14,4 @@ db_packages:
   - rh-mongodb34-syspaths
   - rh-postgresql12-syspaths
   - rh-postgresql12-postgresql-evr
+satellite_ansible_version: 2.9
```
evgeni, line 17: While not hurting now, this is interpreted by YAML as a float, so if we ever have `3.0` here, it will fold it to `3`, which might be not present or mean something completely different. Better quote it.

evgeni:
Overall this reads correctly. I still wonder which Ansible should be used for s-clone, but maybe @mccun934 can shed some light on it.

Seems weird to have to enable the right ansible repo, and then s-clone does enable it again?

jturel:
> Overall this reads correctly. I still wonder which Ansible should be used for s-clone, but maybe @mccun934 can shed some light on it.
> 
> Seems weird to have to enable the right ansible repo, and then s-clone does enable it again?

It's a little wonky but the way it plays out is that it's just making sure that by the time everything is done with the clone process all of the repos are enabled as they would have been if one followed the installation docs, especially since Clone has a step to disable all repos fairly early on (likely to make sure packages are getting pulled from the expected locations).

jturel:
I'm hitting an issue at the end of the process where, after the restoration is performed via foreman-maintain, the installer is run to upgrade to the latest z-stream.

```
TASK [satellite-clone : Run installer upgrade to match latest z-version] *********************************************************************************************************************
Wednesday 09 June 2021  19:13:20 +0000 (0:00:00.228)       0:55:28.751 ********                                                                                                                                                               
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["satellite-installer", "--upgrade", "--disable-system-checks"], "delta": "0:00:07.912125", "end": "2021-06-09 19:13:29.012080", "msg": "non-zero return code", "rc": 1, "start": "202
1-06-09 19:13:21.099955", "stderr": "ERROR: Unrecognised option '--upgrade'\n\nSee: 'satellite-installer --help'", "stderr_lines": ["ERROR: Unrecognised option '--upgrade'", "", "See: 'satellite-installer --help'"], "stdout": "2021-06-09 
19:13:24 [NOTICE] [root] Loading default values from puppet modules...\n2021-06-09 19:13:28 [NOTICE] [root] ... finished", "stdout_lines": ["2021-06-09 19:13:24 [NOTICE] [root] Loading default values from puppet modules...", "2021-06-09 1
9:13:28 [NOTICE] [root] ... finished"]} 
```

Will update to use foreman-maintain for 6.9 and later

roles/satellite-clone/tasks/main.yml
```diff
@@ -279,10 +285,15 @@
...
 
-  - name: Run installer upgrade to match latest z-version
+  - name: Upgrade to latest z-version with satellite-installer
     command: 'satellite-installer --upgrade {{ satellite_upgrade_options | default("") }}'
     environment:
       HOSTNAME: "{{ hostname }}"
+    when: satellite_version is version("6.8", "<=")
+
+  - name: Upgrade to latest z-version with satellite-maintain
+    command: 'satellite-maintain upgrade run --target-version={{ satellite_version }}.z --assumeyes'
```
jturel, line 295: Will this work as far back as 6.5? If so I can remove the above equivalent which used satellite-installer

wbclark: Hi @jturel , great question!

As you saw, `--upgrade` was removed (with https://github.com/theforeman/foreman-installer/pull/541/commits/3c60c745fca4a98cf6e04ebbb3d7b8462373e75a ). In fact, at that point it did nothing except print a notice that `--upgrade` is deprecated since we automatically run any pending db migrations (since https://github.com/theforeman/foreman-installer/pull/470/commits/1cc2378a77c73a544de074d420140f274965ce09 ) was included with Foreman 2.1 (corresponding to Satellite 6.8)

IMO, rather than making the change to use satellite-maintain here with 6.9 and later versions, because satellite-maintain introduces a lot of additional overhead (running various pre-upgrade checks, unlocking packages and updating them, etc) which is intended to simplify the user experience of upgrading for customers, but doesn't have much benefit in this specific context, I would simply make the change to remove `--upgrade` from the task here, but include it in `satellite_upgrade_options` only `when: satellite_version is version("6.7", "<=")`

In other words, logically just running the installer still makes sense at this point, we just need to no longer pass `--upgrade` to it as that was deprecated in Satellite 6.8 and has been removed in Satellite 6.9.

evgeni: Slightly different solution: why don't we add `--upgrade` to `satellite_upgrade_options` for 6.5,6.6,6.7 and leave it out in 6.8+? Removes the need for doing wild `satellite_version` checks here in the playbook as we already `satellite_upgrade_options`?

jturel: Thanks @wbclark & @evgeni - gonna update the vars file to prevent adding back a satellite_version check

jturel: Updated!

evgeni: Nice!

roles/satellite-clone/tasks/main.yml
```diff
@@ -197,19 +197,25 @@
...
-- name: Remove cpdb_done file
+- name: Remove legacy cpdb_done file
   file:
     path: /var/lib/candlepin/cpdb_done
     state: absent
 
+# The location of this file changed in 6.9
+- name: Remove cpdb_done file
+  file:
+    path: /var/lib/candlepin/.puppet-candlepin-cpdb-create-done
```
evgeni, line 216: You could fold this and the legacy path into a list and do a with_items.

Or put the filepath as a variable as for each satellite version there is always just one matching.

Or you leave it like this, as it doesn't hurt ;)

evgeni (commented):
From an Ansible perspective this looks good (I added one comment, but nothing major).

Does a clone run succeed with these changes?

roles/satellite-clone/tasks/main.yml
jturel: I like with_items - will do

jturel:
Updated & squashed!

evgeni (approved):

ekohl (approved):

sthirugn:
yay! @jturel++

PR #333, opened 2018-06-12
Add 6.3 puppet 4 support

johnpmitsch:
Fixes #328

Issue #328, opened 2018-03-14
satellite-clone fails on a satellite 6.3 installed with puppet 4


sthirugn:
I had to add a task to workaround:
```sh
- name: enable puppet4 rpms
  command: subscription-manager repos --enable rhel-7-server-satellite-6.3-puppet4-rpms
  when: enable_repos
```

But the permanent fix will be to check the `puppet-agent` package availability in backup metadata.yml:
```sh
cat /backup/satellite-backup-20180313213434/metadata.yml | grep 'puppet-agent'
```
If the above returned an item (or two - I see `puppet-agent-1` and `puppet-agent-oauth`), then it is puppet4. In other words, a return value of 0 for the above command will indicate puppet4.

sthirugn:
Add Evgeni and Eric as reviewers after implementing this.

johnpmitsch:
we should actually add this to the check_metadata_for_satellite_version module since that is already parsing the yml. This will allow us to return something like
```python
{ satellite_version: 6.3, puppet_version: 4 }
```

sthirugn:
even better, yeah + to use the new ansible module

PR #387, opened 2021-06-08
Remove code related to Puppet version 4

wbclark:
Puppetserver 4 was only present on Satellite 6.3. At present, the oldest
supported Satellite release is Satellite 6.5. So code related to
Puppetserver 4 should be removed.

### DIFF
--- a/library/parse_backup_metadata.py
+++ b/library/parse_backup_metadata.py
@@ -38,15 +38,6 @@ def get_rpm_version(rpms, pattern, hyphen_split=1, version_split=2):
         return False
 
 
-def get_puppet_version(puppet_agent_version, puppet_rpm_version):
-    error_msg = "Puppet version not found"
-    # Only puppet 4+ has puppet-agent rpm
-    if puppet_agent_version:
-        return puppet_agent_version
-    else:
-        raise error_msg
-
-
 def parse_backup_metadata(params):
     with open(params["metadata_path"]) as data_file:
         data = yaml.load(data_file)
@@ -54,9 +45,6 @@ def parse_backup_metadata(params):
     rpm_key = ":rpms" if ":rpms" in data else "rpms"
     rpms = data[rpm_key]
     satellite_version = get_rpm_version(rpms, "^satellite-[\d+].*")
-    puppet_agent_version = get_rpm_version(rpms, "^puppet-agent-[\d+].*", 2, 1)
-    puppet_rpm_version = get_rpm_version(rpms, "^puppet-[\d+].*", 1, 1)
-    puppet_version = get_puppet_version(puppet_agent_version, puppet_rpm_version)
 
     if not satellite_version or satellite_version not in SUPPORTED_VERSIONS:
         msg = "Satellite version is not supported or found. " \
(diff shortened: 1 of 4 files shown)


library/parse_backup_metadata.py
```diff
@@ -56,7 +47,6 @@ def parse_backup_metadata(params):
     satellite_version = get_rpm_version(rpms, "^satellite-[\d+].*")
     puppet_agent_version = get_rpm_version(rpms, "^puppet-agent-[\d+].*", 2, 1)
     puppet_rpm_version = get_rpm_version(rpms, "^puppet-[\d+].*", 1, 1)
```
ekohl, line 49: These variables are now unused

roles/satellite-clone/tasks/main.yml
```diff
@@ -166,15 +160,6 @@
...
   when: clone_foreman_dump_exists and clone_candlepin_dump_exists and (clone_no_foreman_dump_access or clone_no_candlepin_dump_access)
 
-# If there is no puppet user, the extracted config files will not have the correct puppet
-# user/group permissions with puppet 4. Installing puppetserver before extracting config_files
-# fixes this issue.
-- name: Install puppetserver to create puppet user
-  yum:
-    name: puppetserver
-    state: latest
-  when: puppet_version|int == 4
```
ekohl, line 176: Isn't this still true for all Puppetserver versions today?

ekohl (changes requested):
In description I'd just use `Puppet version 4`.

roles/satellite-clone/tasks/main.yml
wbclark: If that's true then there is a bug here, because the above task only runs `when: puppet_version|int == 4`

I wonder what is the historical reason this was needed (can check blame/logs to have a look), as I think the installer should ensure we have correct user/group/permissions as long as we reach that point

wbclark:
Thanks @ekohl . I've updated to use "Puppet" over "puppetserver" in the commit message, and removed the additional unused variables.

Investigating why the puppetserver pre-install was ever needed to begin with and whether it should in fact run with other Satellite/Puppet versions

roles/satellite-clone/tasks/main.yml
wbclark: Git forensics reveals the issue was: https://github.com/RedHatSatellite/satellite-clone/issues/328

The issue report states what was the fix, but does not state what was the original problem

It was resolved in https://github.com/RedHatSatellite/satellite-clone/commit/b0de5e6983b5ba58499cba63136b034301b96960 with a comment at https://github.com/RedHatSatellite/satellite-clone/pull/333#issuecomment-397024596 mentioning permissions issues with puppet, but not further details

In my opinion it's safe to remove this, because this code already doesn't run on any of the currently supported Satellite/Puppet versions

evgeni: Given nobody complained for a while… ;)

ekohl: I can imagine that the puppet user should be present for the backup restore to succeed, but it could very well be that the first installer run rectifies all the permissions.

ekohl (approved):
I don't see how the old code could have worked once we upgraded to Puppet 5 so let's drop it.

evgeni:
Agreed

PR #388, opened 2021-06-15
Disable Smart Proxy Realm plugin

ekohl:
Similar to 5d94c6ee25eccff05c8f894217aed6245e186d34 this disables realm integration on cloning. This is because the new host is usually not enrolled in IPA so the files it tries to read are not present. That causes clones to fail.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1965844

### DIFF
--- a/docs/satellite-clone.md
+++ b/docs/satellite-clone.md
@@ -25,7 +25,7 @@
 
 ### After Cloning ###
   - On the target server, refreshing the manifest will invalidate the source server's manifest.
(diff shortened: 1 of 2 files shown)


jturel:
Looks good & makes sense. Running the automated testing against the change now

ekohl:
I walked through this with the customer but it turns out not to work as expected. While it does allow the step to pass, then it later fails when it runs the actual import with the answers file from the backup. This means it fails on a later step.

PR #392, opened 2021-09-07
Bug fixes and style improvements to Capsule LCE disassociation script

wbclark:
Prior to this commit, the Capsule names are collected but never used.
Removing the collection of Capsule names speeds up execution by avoiding
potentially many redundant calls to `hammer --csv capsule info`.

### DIFF
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -12,8 +12,7 @@ def prepare_hammer_cmd(command)
 end
 
 def run_hammer_cmd(command)
-  command = prepare_hammer_cmd(command)
-  `#{command}`
+  `#{prepare_hammer_cmd(command)}`
 end
 
 def get_ids_from_hammer(command, search = nil)
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -2,6 +2,7 @@
 
 require 'shellwords'
 require 'rubygems'
+require 'json'
 
 @username = "admin"
 @password = "changeme"
@@ -15,17 +16,18 @@ def run_hammer_cmd(command)
   `#{command}`
 end
 
-def get_info_from_hammer(command, column=1)
-  bash_parse = " | grep -v \"Warning:\" | tail -n+2 | awk -F, {'print $#{column}'}"
-  run_hammer_cmd(command + bash_parse).split("\n")
+def get_ids_from_hammer(command, search = nil)
+  JSON.parse(run_hammer_cmd("--output=json " + command + (search ? " --search '#{search}'" : "")).map do |obj|
+    obj["Id"]
+  end
 end
 
 def capsule_lce_args(action, capsule_id, env)
   "--csv capsule content #{action}-lifecycle-environment --id #{capsule_id} --lifecycle-environment-id #{env}"
 end
 
-external_capsule_lifecycle_environments = get_info_from_hammer("--csv capsule list --search 'feature = \"Pulp Node\"'").map do |id|
-  { capsule_id: id, lifecycle_environments: get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}") }
+external_capsule_lifecycle_environments = get_ids_from_hammer("capsule list", search = 'feature = "Pulp Node"').map do |id|
+  { capsule_id: id, lifecycle_environments: get_ids_from_hammer("capsule content lifecycle-environments --id #{id}") }
 end
 
 reverse_commands = external_capsule_lifecycle_environments.map do |association|
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -30,9 +30,10 @@ if external_capsule_ids.empty?
   STDOUT.puts "There are no external capsules to disassociate."
 else
   external_capsule_ids.split("\n").each do |id|
-    lifecycle_environment = get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n")
-    name = get_info_from_hammer("--csv capsule info --id #{id}", 2).chomp
-    external_capsules << {:id => id, :name => name, :lifecycle_environments => lifecycle_environment}
+    external_capsules << {
+      id: id,
+      lifecycle_environments: get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n")
+    }
   end
 
   reverse_commands = []
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -24,30 +24,21 @@ def capsule_lce_args(action, capsule_id, env)
   "--csv capsule content #{action}-lifecycle-environment --id #{capsule_id} --lifecycle-environment-id #{env}"
 end
 
-external_capsules = []
 external_capsule_ids = get_info_from_hammer("--csv capsule list --search 'feature = \"Pulp Node\"'")
-if external_capsule_ids.empty?
-  STDOUT.puts "There are no external capsules to disassociate."
-else
-  external_capsule_ids.split("\n").each do |id|
-    external_capsules << {
-      id: id,
-      lifecycle_environments: get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n")
-    }
-  end
+STDOUT.puts "There are no external capsules to disassociate." if external_capsule_ids.empty?
 
-  reverse_commands = []
-  external_capsules.each do |capsule|
-    capsule[:lifecycle_environments].each do |env|
-      run_hammer_cmd(capsule_lce_args("remove", capsule[:id], env))
-      reverse_commands << prepare_hammer_cmd(capsule_lce_args("add", capsule[:id], env))
-    end
-  end
-
-  STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
-              "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
-              " making sure to replace the credentials with your own."
-  reverse_commands.each do |reverse|
-    STDOUT.puts reverse
-  end
+external_capsules = external_capsule_ids.split("\n").map do |id|
+  { id: id, lifecycle_environments: get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n") }
 end
+
+reverse_commands = external_capsules.map do |capsule|
+  capsule[:lifecycle_environments].map do |env|
+    run_hammer_cmd(capsule_lcs_args("remove", capsule[:id], env))
+    prepare_hammer_cmd(capsule_lce_args("add", capsule[:id], env))
+  end
+end.flatten
+
+STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
+            "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
+            " making sure to replace the credentials with your own." unless reverse_commands.empty?
+reverse_commands.each { |command| STDOUT.puts command }
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -25,8 +25,6 @@ def capsule_lce_args(action, capsule_id, env)
 end
 
 external_capsule_ids = get_info_from_hammer("--csv capsule list --search 'feature = \"Pulp Node\"'")
-STDOUT.puts "There are no external capsules to disassociate." if external_capsule_ids.empty?
-
 external_capsules = external_capsule_ids.split("\n").map do |id|
   { id: id, lifecycle_environments: get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n") }
 end
@@ -38,9 +36,15 @@ reverse_commands = external_capsules.map do |capsule|
   end
 end.flatten
 
-unless reverse_commands.empty?
-  STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
-              "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
-              " making sure to replace the credentials with your own."
+if reverse_commands.empty?
+  STDOUT.puts "There were no associated lifecycle environments to disassociate from external capsules."
+else
+  STDOUT.puts <<~MESSAGE
+    Any external Capsules have been disassociated with any lifecycle environments.
+    This is to avoid any syncing errors with your original Satellite and any
+    interference with existing infrastructure. To reverse these changes, run the
+    following commands, making sure to replace the credentials with your own:
+
+  MESSAGE
   reverse_commands.each { |command| STDOUT.puts command }
 end
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -38,7 +38,9 @@ reverse_commands = external_capsules.map do |capsule|
   end
 end.flatten
 
-STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
-            "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
-            " making sure to replace the credentials with your own." unless reverse_commands.empty?
-reverse_commands.each { |command| STDOUT.puts command }
+unless reverse_commands.empty?
+  STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
+              "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
+              " making sure to replace the credentials with your own."
+  reverse_commands.each { |command| STDOUT.puts command }
+end
--- a/roles/satellite-clone/files/disassociate_capsules.rb
+++ b/roles/satellite-clone/files/disassociate_capsules.rb
@@ -17,18 +17,15 @@ end
 
 def get_info_from_hammer(command, column=1)
   bash_parse = " | grep -v \"Warning:\" | tail -n+2 | awk -F, {'print $#{column}'}"
-  run_hammer_cmd(command + bash_parse)
+  run_hammer_cmd(command + bash_parse).split("\n")
 end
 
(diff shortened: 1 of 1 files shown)


roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -30,9 +30,10 @@ def capsule_lce_args(action, capsule_id, env, lifecycle)
   STDOUT.puts "There are no external capsules to disassociate."
 else
   external_capsule_ids.split("\n").each do |id|
-    lifecycle_environment = get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n")
-    name = get_info_from_hammer("--csv capsule info --id #{id}", 2).chomp
-    external_capsules << {:id => id, :name => name, :lifecycle_environments => lifecycle_environment}
+    external_capsules << {
```
ekohl, line 33: I'm just wondering if it should instead do something like:
```ruby
external_capsules = external_capsule_ids.split("\n").map do |id|
  {
    ...
  }
end

**Review by ekohl — COMMENTED** [review]
http://bash.org/?240849
> \<Patrician|Away> what does your robot do, sam
> \<bovril> it collects data about the surrounding environment, then discards it and drives into walls

File: roles/satellite-clone/files/disassociate_capsules.rb (hunk 1 above)
Comment (wbclark) [reply roles/satellite-clone/files/disassociate_capsules.rb]: Nice idea

I updated the style, please take a look and let me know what you think

File: roles/satellite-clone/files/disassociate_capsules.rb [hunk 2]
```diff
@@ -24,30 +24,21 @@ def capsule_lce_args(action, capsule_id, env)
...
-if external_capsule_ids.empty?
-  STDOUT.puts "There are no external capsules to disassociate."
-else
-  external_capsule_ids.split("\n").each do |id|
-    external_capsules << {
-      id: id,
-      lifecycle_environments: get_info_from_hammer("--csv capsule content lifecycle-environments --id #{id}").split("\n")
-    }
-  end
+STDOUT.puts "There are no external capsules to disassociate." if external_capsule_ids.empty?
```
ekohl, line 28: Removing the if/else this way does change the behavior. Now it will also print the long message at the end of the script if it's empty.

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -24,30 +24,21 @@ def capsule_lce_args(action, capsule_id, env)
   "--csv capsule content #{action}-lifecycle-environment --id #{capsule_id} --lifecycle-environment-id #{env}"
 end
 
-external_capsules = []
 external_capsule_ids = get_info_from_hammer("--csv capsule list --search 'feature = \"Pulp Node\"'")
```
ekohl, line 27: Unrelated, but this will fail on 6.10 where the `Pulp Node` feature is no longer present.

ekohl (changes requested):

roles/satellite-clone/files/disassociate_capsules.rb
wbclark: Note the inline conditional which I added here:

```ruby
STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
            "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
            " making sure to replace the credentials with your own." unless reverse_commands.empty?
```

But I had some concern about readability when I wrote it this way. Would it be better style to wrap that in an if, since the string is split across several lines?

ekohl: I think that the `if/else` nesting that was there was easier to follow.

wbclark: I updated it with a block `unless reverse_commands.empty?` conditional for the final output.

Technically this does fix an issue that existed previously where the long final output message would be displayed, with no following reverse_commands, if there were external capsules but the LCEs were already disassociated (hence `external_capsule_ids` would be populated while `reverse_commands` would be empty). That is kind of an edge case but the logic didn't quite seem right and `unless reverse_commands.empty?` more accurately captures what it should really care about.

A side effect of that fix is that I opted to leave the `external_capsules = ...` and `reverse_commands = ...` definition blocks at the top level rather than under an `if/else`, since they will wind up empty anyway if the arrays they are built from are also empty and `if external_capsule_ids.empty?` was never really the correct conditional for every bit of code underneath it.

Let me know what you think, thanks!

wbclark: Looking at it some more, I think ` STDOUT.puts "There are no external capsules to disassociate." if external_capsule_ids.empty?` also belongs in the same block down below for similar reasons. I think the correct form should be:

```ruby
if reverse_commands.empty?
  STDOUT.puts "There were no Lifecycle Environments to disassociate from external capsules."
else
  STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
              "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
              " making sure to replace the credentials with your own."
   reverse_commands.each { |command| STDOUT.puts command }
 end
 ```

File: roles/satellite-clone/files/disassociate_capsules.rb (hunk 3 above)
Comment (wbclark) [reply roles/satellite-clone/files/disassociate_capsules.rb]: At the time this runs, the hostname changeover has already been performed. So I think this can be solved with:

```ruby
require 'socket'
internal_capsule = Socket.gethostname
external_capsule_lifecycle_environments = get_info_from_hammer("--csv capsule list --search 'name != \"#{internal_capsule}\"'").split("\n").map do |id|
  {
    ...
  }
end
```

I am not sure if it's possible that the internal capsule could be renamed to no longer match Satellite's hostname.

Searching by id is not an option unfortunately:

```shell
# hammer --csv capsule list --search 'id != 1'
400 Bad Request
  Field 'id' not recognized for searching!
```

wbclark:
Force pushed to correct typo in commit message

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -17,36 +17,33 @@ def run_hammer_cmd(command)
 
 def get_info_from_hammer(command, column=1)
   bash_parse = " | grep -v \"Warning:\" | tail -n+2 | awk -F, {'print $#{column}'}"
```
ekohl, line 19: I only now read this and I already wondered why it wasn't needed to use `--columns` in the hammer command. You can drop the tail command by using `--no-headers` and I think to avoid `awk` you can use `--columns` (but I don't see it right now in `--help`). However, separate issue.

wbclark: I know it's beyond the original scope of this PR, but I found an elegant way to solve it with just a few lines of change, using `hammer --output=json ...` instead of `hammer --csv ...`

BTW, I appreciate your insights and observations in these reviews -- they have been very valuable to me.

wbclark: Another note about it, `--no-headers` wasn't necessary when using `hammer --output=json`

But looking over my work, I realized I did remove the piece that filters warnings. I'm not immediately sure whether those are stdout or stderr with hammer (I would assume the latter, but haven't checked) but without filtering, they would definitely break the JSON parsing :/

This all makes me wonder whether it's better to use the API directly to bypass the shell, or if it's better to rely on hammer as the API client in case the API changes in the future (although hammer itself can also change, so I don't know)

Also there should probably be some rescue for `JSON::ParserError`

If I don't think of an obvious solution soon, I'll probably split that piece off into a separate PR

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -11,42 +12,39 @@ def prepare_hammer_cmd(command)
...
-  command = prepare_hammer_cmd(command)
-  `#{command}`
+  `#{prepare_hammer_cmd(command)}`
 end
 
-def get_info_from_hammer(command, column=1)
-  bash_parse = " | grep -v \"Warning:\" | tail -n+2 | awk -F, {'print $#{column}'}"
-  run_hammer_cmd(command + bash_parse)
+def get_ids_from_hammer(command, search = nil)
+  JSON.parse(run_hammer_cmd("--output=json " + command + (search ? " --search #{search}" : "")).map do |obj|
```
ekohl, line 19: You should quote search since it contains spaces:
```suggestion
  JSON.parse(run_hammer_cmd("--output=json " + command + (search ? " --search '#{search}'" : "")).map do |obj|
```

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -11,42 +12,39 @@ def prepare_hammer_cmd(command)
...
-  reverse_commands = []
-  external_capsules.each do |capsule|
-    capsule[:lifecycle_environments].each do |env|
-      run_hammer_cmd(capsule_lce_args("remove", capsule[:id], env))
-      reverse_commands << prepare_hammer_cmd(capsule_lce_args("add", capsule[:id], env))
-    end
+reverse_commands = external_capsule_lifecycle_environments.map do |association|
+  association[:lifecycle_environments].map do |env|
+    run_hammer_cmd(capsule_lcs_args("remove", association[:capsule_id], env))
+    prepare_hammer_cmd(capsule_lce_args("add", association[:capsule_id], env))
```
ekohl, line 35: I would keep this in the `else` part. While it's probably cheap, it's still something you don't need to do at all if `external_capsule_lifecycle_environments` is empty. The previous code wasn't poorly structured. There's certainly benefits to being able to decide very early on that you can abort execution.

There's a certain risk in blowing up the scope of a patch. While I certainly understand the desire to restructure things, it's important to find a good balance. The more you change, the more you can break.

wbclark: I got curious to know just how cheap or expensive is that loop when `external_capsule_lifeycycle_environments.empty?` or `external_capsule_lifecycle_environments.all? { |association| association[:lifecycle_environments].empty? }`

So I did a test:

```
irb(main):001:1* def timed_loop(somearray)
irb(main):002:1*   start = Time.now
irb(main):003:1*   
irb(main):004:2*   somearray.map do |elem|
irb(main):005:3*     elem[:some_key].map do |subelem|
irb(main):006:3*       "slept for #{sleep(60)} seconds"
irb(main):007:2*     end
irb(main):008:1*   end.flatten
irb(main):009:1*   
irb(main):010:1*   finish = Time.now
irb(main):011:1*   finish - start
irb(main):012:0> end
=> :timed_loop
irb(main):013:0> timed_loop([])
=> 2.259e-06
irb(main):014:0> timed_loop([{ some_key: [] }])
=> 5.11e-06
```

So it's on the order of a few microseconds. If I instead test something like 1000 "capsules" that each have no LCE assignments:

```
irb(main):036:1* myarray = 1000.times.map do |count|
irb(main):037:1*   { id: count, some_key: [] }
irb(main):038:0> end
=> [{:id=>0, :some_key=>[]}, {:id=>1, :some_key=>[]}, {:id=>2, :some_key=>[]}, {:id=>3, :some_key=>[]}, {:id=>4, :some_key=>[]}, {:id=>5, :some_key=>[]}, {:id=>6, :some_key=...
irb(main):039:0> timed_loop(myarray)
=> 0.000413076
```

Then it gets up to almost half a millisecond on my machine.

I realize the execution time isn't the point (we are on the same page about that) but I was just curious to know how it behaves :)

wbclark: But as we can see, since `"slept for #{sleep(60)} seconds"` never gets called in the above tests, in an implicit sense the interior of the loop is already within that if/else (speaking about the logic), the only difference being that we get `reverse_commands = []` instead of having it be undefined.

And I rather like that, because using `reverse_commands.empty?` seems like the best possible test for whether we need to print some output containing the reversal commands.

wbclark: This actually fixes a bug that existed previously, where we would try to print reversal commands and the full message about disassociation, when external capsules existed but their LCEs were already fully disassociated.

Fixing that bug otherwise would require changing the condition to `external_capsule_lifecycle_environments.all? { |association| association[:lifecycle_environments].empty? }` which is strictly worse because it still requires looping over `external_capsule_lifecycle_environments` once when they are all empty (same as my approach) but also requires looping over them twice (once for the conditional, once again to do the disassociations and populate `reverse_commands`) when they are populated, whereas my solution requires that only once in all cases.

roles/satellite-clone/files/disassociate_capsules.rb
wbclark: It's quoted when it's assigned a value, when the function is called at https://github.com/RedHatSatellite/satellite-clone/pull/392/files#diff-e156d7efb3320dd71eaac5bb78d7f6f2edd77b5aa05b767fbfdad959859fb463R28

And that seems to work fine:

```ruby
irb(main):043:0> search = 'feature = "Pulp Node"'
=> "feature = \"Pulp Node\""
irb(main):044:0> "--output=json " + "hammer capsule list" + (search ? " --search #{search}" : "")
=> "--output=json hammer capsule list --search feature = \"Pulp Node\""
```

(It also worked when I live tested these functions on Satellite server)

wbclark: Just kidding, I actually had this in my test environment but not here. I will correct the issue shortly.

Good catch. Thank you!

wbclark: Updated, it should be correct now:

```ruby
irb(main):045:0> search = 'feature = "Pulp Node"'
=> "feature = \"Pulp Node\""
irb(main):046:0> "--output=json " + "hammer capsule list" + (search ? " --search '#{search}'" : "")
=> "--output=json hammer capsule list --search 'feature = \"Pulp Node\"'"
```

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -11,42 +12,39 @@ def prepare_hammer_cmd(command)
...
-  STDOUT.puts "All Capsules are unassociated with any lifecycle environments. This is to avoid any syncing errors with your original Satellite " \
-              "and any interference with existing infrastructure. To reverse these changes, run the following commands," \
-              " making sure to replace the credentials with your own."
-  reverse_commands.each do |reverse|
-    STDOUT.puts reverse
-  end
+if reverse_commands.empty?
+  STDOUT.puts "There were no associated lifecycle environments to disassociate from external capsules."
+else
+  STDOUT.puts <<~MESSAGE
```
evgeni, line 42: this doesn't work on Ruby 2.0 (which we still have to support), as it doesn't know "squiggly heredoc" (2.3+ feature)

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -11,42 +12,39 @@ def prepare_hammer_cmd(command)
...
-  `#{command}`
+  `#{prepare_hammer_cmd(command)}`
 end
 
-def get_info_from_hammer(command, column=1)
-  bash_parse = " | grep -v \"Warning:\" | tail -n+2 | awk -F, {'print $#{column}'}"
-  run_hammer_cmd(command + bash_parse)
+def get_ids_from_hammer(command, search = nil)
+  JSON.parse(run_hammer_cmd("--output=json " + command + (search ? " --search '#{search}'" : "")).map do |obj|
+    obj["Id"]
```
evgeni, line 20: this is funny (well, actually not) but I recall this sometimes being `Id` and sometimes `ID`, depending on the hammer version…

evgeni (changes requested):
Let me be honest, this PR creeped out any possible scope by at least 9000. But there is also a lot of valuable change here.

Looking at the individual commits:
* Do not collect Capsule names in disassociation script :heavy_check_mark: 
* Style updates to disassociate_capsules.rb :heavy_check_mark: 
* Re-instate block conditional vs. in line for Capsule reverse commands :heavy_check_mark: 
* Move output messages into common if/else block :heavy_check_mark: (the three should probably be squashed to one, tho, and needs fixing the squiggly thingy dingy)
* Remove redundant variables and rename remaining for better accuracy :no_entry_sign: -- I find this one very hard to read, and would love to look it it without the rest of the changes
* Always split get_info_from_hammer output :no_entry_sign: -- technically obsolete by the next one
* Retrieve only ids for Capsules and LCEs :no_entry_sign: -- I smell potential breakage here and would like to review/test it on its own
* simplify run_hammer_cmd :heavy_check_mark: 

If you could split out the "easy" ones and fix up that heredoc, we could merge them, and then look at the renaming and json changes in individual PRs.

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -11,42 +12,39 @@ def prepare_hammer_cmd(command)
 end
 
 def run_hammer_cmd(command)
-  command = prepare_hammer_cmd(command)
-  `#{command}`
+  `#{prepare_hammer_cmd(command)}`
 end
 
-def get_info_from_hammer(command, column=1)
-  bash_parse = " | grep -v \"Warning:\" | tail -n+2 | awk -F, {'print $#{column}'}"
```
evgeni, line 19: while the pipes and the greps and the tails are awkward, the new code will fail if hammer outputs e.g. this:

```
# hammer --output json proxy list
NOTE: Gem.gunzip is deprecated; use Gem::Util.gunzip instead. It will be removed on or after 2018-12-01.
Gem.gunzip called from /usr/lib/ruby/vendor_ruby/unicode/display_width/index.rb:5.
[
  {
    "Id": 1,
    "Name": "ubuntu2004-foreman-3-0.yatsu.example.com",
    "URL": "https://ubuntu2004-foreman-3-0.yatsu.example.com:8443",
    "Features": [
      "Puppet CA",
      "Puppet",
      "Logs",
      "Registration"
    ]
  }
]
```

The hammer call should at least gain a `2>/dev/null` to drop stderr.

roles/satellite-clone/files/disassociate_capsules.rb
evgeni: https://github.com/Katello/hammer-cli-katello/commit/416e7f780acc3a958e8c9e0ea0d4e565766002f3

roles/satellite-clone/files/disassociate_capsules.rb
wbclark: :+1: thanks for the reminder to do this

roles/satellite-clone/files/disassociate_capsules.rb
wbclark: Oof. In what cases do we still support Ruby 2.0?

ekohl: EL7

wbclark: Ah right, because this script uses system ruby. TY

PR #393, opened 2021-09-07
Always use katello:reimport rake, never katello:reindex

wbclark:
`foreman-rake katello:reindex` was previously replaced with
`foreman-rake katello:reimport`. Currently supported versions use
reimport only, so variables tracking which command must be used are dropped.
The variable previously named `run_katello_reindex` is also renamed to
be accurate to the currently supported Satellite versions.

### DIFF
--- a/satellite-clone-vars.sample.yml
+++ b/satellite-clone-vars.sample.yml
@@ -17,9 +17,9 @@
 # Disable firewall to setup Satellite for testing purposes. (defaults to false)
 #disable_firewall: false
 
-# Run katello reindex after the Satellite install. (defaults to false)
+# Run `foreman-rake katello:reimport` after the Satellite install. (defaults to false)
 # Note: If you are planning to clone your Satellite and immediately upgrade, you may choose to skip this step.
-#run_katello_reindex: false
+#run_katello_reimport: false
 
 # Run several pre-install checks (defaults to true)
 #run_pre_install_check: true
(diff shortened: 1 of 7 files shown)


satellite-clone-vars.sample.yml
```diff
@@ -17,9 +17,9 @@
 # Disable firewall to setup Satellite for testing purposes. (defaults to false)
 #disable_firewall: false
 
-# Run katello reindex after the Satellite install. (defaults to false)
+# Run `foreman-rake katello:reimport` after the Satellite install. (defaults to false)
 # Note: If you are planning to clone your Satellite and immediately upgrade, you may choose to skip this step.
-#run_katello_reindex: false
+#run_katello_reimport: false
```
ekohl, line 22: Users can have this config file in `/etc`. Do we need to worry about those things? I really don't know what our support policy is in this repo.

ekohl (commented):
:+1: for getting rid of the `verify_rake_task` variable.

chris1984:
> /lgtm

Aren't you on insights now, /shoo

evgeni:
I've dropped ` roles/satellite-clone/vars/satellite_6.5.yml ` over in #391, mind rebasing? :)

wbclark:
> I've dropped `roles/satellite-clone/vars/satellite_6.5.yml` over in #391, mind rebasing? :)

Sure thing, thanks

satellite-clone-vars.sample.yml
evgeni: I think this is fine?

cc @mccun934

evgeni:
Absolutely happy with the change. The only thing I wonder about is the backwards compatibility, which we're breaking by renaming that variable.

Personally, I am happy to break, as I expect users to re-visit their config anyways, as they are also cloning a new version of Satellite, but you never know ;)

evgeni (approved):

PR #395, opened 2021-09-14
use answers file variable

evgeni:


ekohl (approved):
While a good bugfix in itself, it doesn't fix the first part of https://bugzilla.redhat.com/show_bug.cgi?id=1965844 which is that it doesn't read the scenario answers path from the scenario definition. Are you looking at that as well?

evgeni:
> While a good bugfix in itself, it doesn't fix the first part of https://bugzilla.redhat.com/show_bug.cgi?id=1965844 which is that it doesn't read the scenario answers path from the scenario definition. Are you looking at that as well?

No, I just want to pass `satellite_answers_path: katello-answers.yml` :rofl:

PR #397, opened 2021-09-15
Satellite 6.10 support

evgeni:


evgeni:
I intentionally didn't add it to the `README.md` yet, as it's incomplete.

ekohl (commented):
`Pulp Node` usage also needs fixing since that's Pulp 2: https://github.com/RedHatSatellite/satellite-clone/search?q=pulp+node

wbclark:
Thanks @evgeni . Would you consider getting #392 in first (and if so, should this be rebased?)

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -24,8 +24,11 @@ def capsule_lce_args(action, capsule_id, env)
   "--csv capsule content #{action}-lifecycle-environment --id #{capsule_id} --lifecycle-environment-id #{env}"
 end
 
+# the correct way is to look at hosts with the Pulpcore feature and the mirror capability
+# but hammer doesn't support that
+internal_capsule = ARGV[0]
 external_capsules = []
-external_capsule_ids = get_info_from_hammer("--csv capsule list --search 'feature = \"Pulp Node\"'")
+external_capsule_ids = get_info_from_hammer("--csv capsule list --search '(feature = \"Pulp Node\" or feature = \"Pulpcore\") and name != \"#{internal_capsule}\"'")
```
wbclark, line 31:
```suggestion
external_capsule_ids = get_info_from_hammer("--csv capsule list --search '(name != \"#{internal_capsule}\"'")
```

I think this is simpler and works just as well; downstream does not support disabling Pulp on Capsules

roles/satellite-clone/files/disassociate_capsules.rb
```diff
@@ -24,8 +24,11 @@ def capsule_lce_args(action, capsule_id, env)
   "--csv capsule content #{action}-lifecycle-environment --id #{capsule_id} --lifecycle-environment-id #{env}"
 end
 
+# the correct way is to look at hosts with the Pulpcore feature and the mirror capability
+# but hammer doesn't support that
+internal_capsule = ARGV[0]
```
wbclark, line 29: I was thinking this might be `internal_capsule = Socket.gethostname` since the hostname change has already been performed when this script is run.

But your approach of relying on the ansible variable value seems very nice and consistent. :+1:

evgeni: I have bad experience with Python's `socket.getfqdn()`, so I'll pass ;)

evgeni:
> Thanks @evgeni . Would you consider getting #392 in first (and if so, should this be rebased?)

Yeah, sounds good.

roles/satellite-clone/files/disassociate_capsules.rb
evgeni: That's probably true.

One thing that bothers me: you can update the name of a capsule, and nothing will break… or should, with the code change here, it will :D

evgeni: We could do `url !~ https://#{internal_capsule}` instead, but that also doesn't feel right

ekohl: I would prefer not to build that in. The whole concept of a Smart Proxy on the same host is IMHO a Katelloism I'd love to kill. As we're looking at other architectures, we may at some point support a `satellite.example.com` without a Capsule and then a `content.example.com` which is your primary Capsule. If you start relying only on hostnames it becomes even more obscure that you really do want Smart Proxies that have content.

> I think this is simpler and works just as well; downstream does not support disabling Pulp on Capsules

It's not supported today, but your suggestion makes it harder to support it tomorrow.

We should really file an RFE for Hammer to expose the capability. The API does.

evgeni:
I've dropped the capsule part from this PR, making it strictly Satellite only. This makes it:
- not conflict with #392, so we can merge them independently
- easier for us to go and get a proper solution for the capsule problem

evgeni:
cc @mccun934 @sthirugn

ekohl (commented):
How do we deal with that this is only partial support? Do we document this somehow?

evgeni:
I'd just not document 6.10 being supported at all (in the readme) yet

ekohl:
The line `SUPPORTED_VERSIONS` could be a bit confusing. Perhaps the PR title should also reflect that it's partial?

PR #399, opened 2021-09-20
detect answers file path automatically

evgeni:
includes #398, for obvious reasons ;)


ekohl (approved):
This at least partly addresses https://bugzilla.redhat.com/show_bug.cgi?id=1965844.

PR #400, opened 2021-09-20
define installer, scenario and package in defaults, not as vars for every version

evgeni:


ekohl (approved):
Does it make sense to also define `satellite_installer_options` and `satellite_upgrade_options` in defaults (and override in 6.6/6.7)?

evgeni:
probably, so far I've only vent for the obvious which were the same across all versions. I'll keep that in mind for the next round (which will come, for sure)

PR #401, opened 2021-10-01
Capsule 6.10 support

evgeni:
this is the capsule part of #397, now in an own PR, so that it doesn't clash with other enhancements while we discuss the correct path forward.

to recap the problem:
- with Pulp3, both the internal and the external capsules expose the `Pulpcore` feature
- this makes it hard to identify only external capsules to disassociate them
- the most correct approach would be to look at the `Mirror` capability of the `Pulpcore` feature, but this is not exposed in Hammer right now.


evgeni:
@ekohl one thought that I had: we can "work around" the issue for now by searching for proxies that have the Pulpcore *and* the Container_Gateway features, as I think CG is only ever enabled on external proxies?

ekohl:
That sounds like a less ugly workaround than hostname. I'd :+1: that. I also think that for all Satellite versions this would be true. So it'd be `Pulp Node or (Pulpcore and CG)`.

evgeni:
technically, we don't even need to look for `Pulpcore`, as that will be true for all *Satellite* deployments anyways.

ekohl:
While that's true today, I'd still prefer to have it in. There's been talk about REX-only Capsules and it'd be great if we one day could support that.

evgeni:
> While that's true today, I'd still prefer to have it in. There's been talk about REX-only Capsules and it'd be great if we one day could support that.

okay, what's the correct `HAVING` SQL statement is for that? ;-)

PR #409, opened 2021-11-05
allow using katello instead of satellite

evgeni:


evgeni:
The alternative would be to accept `SATELLITE_CLONE_ALLOW_KATELLO` as the Satellite *version*, so I can `export SATELLITE_CLONE_ALLOW_KATELLO=6.10` and it behaves like it's a 6.10.

Then we could even call it `SATELLITE_CLONE_FORCE_VERSION` and "shortcut" the whole Katello path?

library/parse_backup_metadata.py
```diff
@@ -48,6 +49,10 @@ def parse_backup_metadata(params):
     puppetserver_present = bool(find_rpm(rpms, "^puppetserver-[\d+].*"))
     qpidd_present = bool(find_rpm(rpms, "^qpid-cpp-server-[\d+].*"))
 
+    if os.getenv('SATELLITE_CLONE_ALLOW_KATELLO'):
```
ekohl, line 52: Should it automatically fall back to this if `satellite_version` is empty (as it would be in upstream)?
```suggestion
    if not satellite_version:
```

ehelms:
> Then we could even call it `SATELLITE_CLONE_FORCE_VERSION` and "shortcut" the whole Katello path?

This makes sense. They are both hacks so I am cool with whatever you are finding helps testing.

ekohl (commented):
Perhaps more of a longer term thing, but should we instead make katello versions leading?

library/parse_backup_metadata.py
evgeni: Right now, this is `satellite-clone`, and it has safeguards to verify it actually is a Satellite it's cloning.
Katello is not Satellite, and I'd prefer to keep this for now (and probably never support it in this form, but teach foreman-maintain how to clone both kinds of setups).

The env var makes it explicitly "user opt-in"

evgeni:
> > Then we could even call it `SATELLITE_CLONE_FORCE_VERSION` and "shortcut" the whole Katello path?
> 
> This makes sense. They are both hacks so I am cool with whatever you are finding helps testing.

From a "as least invasive change as possible" for sat-clone, I'd prefer `SATELLITE_CLONE_FORCE_VERSION`.

ekohl (approved):

ehelms (approved):

ekohl:
> From a "as least invasive change as possible" for sat-clone, I'd prefer `SATELLITE_CLONE_FORCE_VERSION`.

:+1: to this

Issue #402, opened 2021-10-01
Satellite 6.10+ needs a new way of reseting pulp


evgeni:
6.10 uses Pulp 3, which has a different database layout.

probably something like this, instead of all the mongodb stuff, will work:
- drop pulpcore db
- create pulpcore db
- migrate pulpcore db

evgeni:
see https://docs.pulpproject.org/pulpcore/workflows/reclaim-disk-space.html

PR #414, opened 2022-02-08
reset pulpcore data for Satellite 6.10+

evgeni:
Fixes: #402

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -319,12 +319,14 @@
 
   - include_tasks: reset_pulp_data.yml
     when:
+      - reset_pulp_data
       - not clone_pulp_data_exists
       - not online_backup
       - satellite_version in ["6.6", "6.7", "6.8", "6.9"]
 
   - include_tasks: reset_pulpcore_data.yml
     when:
+      - reset_pulp_data
       - not clone_pulp_data_exists
       - not online_backup
       - satellite_version not in ["6.6", "6.7", "6.8", "6.9"]
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -322,14 +322,14 @@
       - reset_pulp_data
       - not clone_pulp_data_exists
       - not online_backup
-      - satellite_version in ["6.6", "6.7", "6.8", "6.9"]
+      - has_pulp2
 
   - include_tasks: reset_pulpcore_data.yml
     when:
       - reset_pulp_data
       - not clone_pulp_data_exists
       - not online_backup
-      - satellite_version not in ["6.6", "6.7", "6.8", "6.9"]
+      - not has_pulp2
 
   - name: Wait 60 seconds for services to be fully up
     wait_for: timeout=60
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -323,6 +323,12 @@
       - not online_backup
       - satellite_version in ["6.6", "6.7", "6.8", "6.9"]
 
+  - include_tasks: reset_pulpcore_data.yml
+    when:
+      - not clone_pulp_data_exists
+      - not online_backup
+      - satellite_version not in ["6.6", "6.7", "6.8", "6.9"]
+
   - name: Wait 60 seconds for services to be fully up
     wait_for: timeout=60
 
--- a/roles/satellite-clone/tasks/reset_pulpcore_data.yml
+++ b/roles/satellite-clone/tasks/reset_pulpcore_data.yml
@@ -1,0 +1,22 @@
+- name: reset pulpcore data
+  command: echo "Katello::Rpm.all.destroy_all; Katello::Erratum.all.destroy_all; Katello::PackageGroup.all.destroy_all; Katello::DockerManifest.all.destroy_all; Katello::DockerTag.all.destroy_all; Katello::AnsibleCollection.all.destroy_all" | foreman-rake console
+
+- name: reset 7.0+ specific pulpcore data
+  command: echo "Katello::GenericContentUnit.all.destroy_all" | foreman-rake console
+  when:
+    - satellite_version not in ["6.10"]
+
+- name: stop all services but the database
+  command: foreman-maintain service stop --exclude postgresql
+
+- name: flush pulpcore db (reset pulp data)
+  command: sudo -u pulp PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager flush --noinput
+
+- name: migrate pulpcore db (reset pulp data)
+  command: sudo -u pulp PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager migrate --noinput
+
+- name: reset pulpcore admin password (reset pulp data)
+  command: sudo -u pulp PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager reset-admin-password --random
+
+- name: start services
+  command: foreman-maintain service start
--- a/roles/satellite-clone/defaults/main.yml
+++ b/roles/satellite-clone/defaults/main.yml
@@ -17,3 +17,4 @@ skip_satellite_rpm_check: False
 overwrite_etc_hosts: True
 check_networking_interfaces: True
 selinux_packages: []
+reset_pulp_data: False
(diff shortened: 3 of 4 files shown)


roles/satellite-clone/tasks/reset_pulp_data.yml
```diff
@@ -1,17 +1,58 @@
...
-- name: reset pulp repo importers
+- name: reset pulp2 repo importers
   command: 'mongo pulp_database --eval "db.repo_importers.update({\"scratchpad.repomd_revision\": {$exists: true}}, {$set: {\"scratchpad.repomd_revision\": null}}, {multi: true});"'
+  when:
+    - satellite_version in ["6.6", "6.7", "6.8", "6.9"]
 
-- name: reset pulp data
-  command: echo "Katello::Rpm.all.destroy_all; Katello::Erratum.all.destroy_all; Katello::PackageGroup.all.destroy_all; Katello::PuppetModule.all.destroy_all; Katello::DockerManifest.all.destroy_all; Katello::DockerTag.all.destroy_all" | foreman-rake console
+- name: reset pulp data in Katello
+  command: echo "Katello::Rpm.all.destroy_all; Katello::Erratum.all.destroy_all; Katello::PackageGroup.all.destroy_all; Katello::DockerManifest.all.destroy_all; Katello::DockerTag.all.destroy_all" | foreman-rake console
```
evgeni, line 12: I realized we never destroyed `Katello::Srpm`, is this something we want to add? @jlsherrill?

roles/satellite-clone/tasks/reset_pulp_data.yml
```diff
@@ -1,17 +1,58 @@
...
-- name: migrate pulp db (reset pulp data)
+- name: reset pulpcore-specific data in Katello
+  command: echo "Katello::AnsibleCollection.all.destroy_all" | foreman-rake console
+  when:
+    - satellite_version in ["6.10", "7.0"]
+
+- name: reset pulpcore-specific data in Katello
+  command: echo "Katello::GenericContentUnit.all.destroy_all" | foreman-rake console
+  when:
+    - satellite_version in ["7.0"]
```
evgeni, line 27: we don't yet support 7.0, but I thought I'd go ahead and add this one, as we know it will come :)

roles/satellite-clone/tasks/reset_pulp_data.yml
```diff
@@ -1,17 +1,58 @@
...
+
+- name: drop pulpcore db
+  command: sudo -u postgres dropdb pulpcore
+  when:
+    - satellite_version in ["6.10", "7.0"]
+
+- name: recreate pulpcore db
+  command: sudo -u postgres createdb --owner=pulp --encoding=UTF8 --locale=en_US.utf8 --template=template0 pulpcore
+  when:
+    - satellite_version in ["6.10", "7.0"]
```
ekohl, line 45: Unless it's a remote DB?

roles/satellite-clone/tasks/reset_pulp_data.yml
```diff
@@ -1,17 +1,58 @@
...
+
+- name: reset pulpcore-specific data in Katello
+  command: echo "Katello::GenericContentUnit.all.destroy_all" | foreman-rake console
+  when:
+    - satellite_version in ["7.0"]
+
+- name: stop all services but the databases
+  command: foreman-maintain service stop --exclude postgresql,rh-mongodb34-mongod
```
ekohl, line 30: Wouldn't it be safer if the services were stopped before you destroy content in Katello? Or do you need foreman-tasks to be running?

ekohl (commented):
I wonder if it makes sense to write a separate file to drop Pulp 2 and another to drop Pulp 3. Perhaps it would be easier on the conditionals.

roles/satellite-clone/tasks/reset_pulp_data.yml
evgeni: yeah, I wasn't sure how to handle that…

roles/satellite-clone/tasks/reset_pulp_data.yml
evgeni: No, I don't think I need anything running, and yes I think it would be safer to stop services the first thing in this play, but I didn't want to refactor *all* of it just now.

evgeni:
> I wonder if it makes sense to write a separate file to drop Pulp 2 and another to drop Pulp 3. Perhaps it would be easier on the conditionals.

Maybe, yeah.

roles/satellite-clone/tasks/reset_pulp_data.yml
evgeni: `pulpcore-manager flush` explicitly documents:
> Removes ALL DATA from the database, including data added during migrations. Does not achieve a "fresh install" state.

:woman_shrugging:

evgeni: on the other hand… let's try that!

roles/satellite-clone/tasks/reset_pulp_data.yml
```diff
@@ -1,17 +1,53 @@
...
+
+- name: migrate pulp2 db (reset pulp data)
   command: sudo -u apache pulp-manage-db
+  when:
+    - satellite_version in ["6.6", "6.7", "6.8", "6.9"]
+
+- name: flush pulpcore db (reset pulp data)
+  command: sudo -u pulp PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager flush --noinput
```
ekohl, line 38: Isn't there some built in magic in Ansible to do sudo? Or would that break things?

evgeni: It can, but I am not sure if it will pass the env var correctly (sudo drops the env, by default) and I kinda didn't want to experiment too much

evgeni:
> I wonder if it makes sense to write a separate file to drop Pulp 2 and another to drop Pulp 3. Perhaps it would be easier on the conditionals.

implemented now, better readable indeed

roles/satellite-clone/tasks/main.yml
```diff
@@ -319,10 +319,18 @@
...
       - not clone_pulp_data_exists
       - not online_backup
       - satellite_version in ["6.6", "6.7", "6.8", "6.9"]
 
+  - include_tasks: reset_pulpcore_data.yml
+    when:
+      - reset_pulp_data
+      - not clone_pulp_data_exists
+      - not online_backup
+      - satellite_version in ["6.10"]
```
ekohl, line 332: Elsewhere you mention 7.0 specific actions, but this would prevent that. Should this be ready for the future and say
```suggestion
      - satellite_version not in ["6.6", "6.7", "6.8", "6.9"]
```

ekohl (commented):
Mostly looks good to me :+1:

roles/satellite-clone/tasks/main.yml
```diff
@@ -319,10 +319,18 @@
 
   - include_tasks: reset_pulp_data.yml
     when:
+      - reset_pulp_data
       - not clone_pulp_data_exists
       - not online_backup
```
evgeni, line 324: technically, I think these two could go, as we now pass the decision to the user

roles/satellite-clone/tasks/reset_pulpcore_data.yml
```diff
@@ -0,0 +1,22 @@
+- name: reset pulpcore data
+  command: echo "Katello::Rpm.all.destroy_all; Katello::Erratum.all.destroy_all; Katello::PackageGroup.all.destroy_all; Katello::DockerManifest.all.destroy_all; Katello::DockerTag.all.destroy_all; Katello::AnsibleCollection.all.destroy_all" | foreman-rake console
+
+- name: reset 7.0+ specific pulpcore data
+  command: echo "Katello::GenericContentUnit.all.destroy_all" | foreman-rake console
+  when:
+    - satellite_version in ["7.0"]
```
ekohl, line 7: If it's 7.0+, should this be:
```suggestion
    - satellite_version not in ["6.10"]
```

evgeni: indeed.

ekohl (approved):
This is good to merge, but should this be rebased now #416 is merged and a fact is available?

ekohl (approved):
Thanks!

ehelms (approved):

PR #415, opened 2022-03-04
ensure pulp user exists and owns /var/lib/pulp for 6.10+

evgeni:
in the case users utilize rsync instead of a backup with pulp data, they
might end up with /var/lib/pulp not being owned by pulp, let's fix that!

### DIFF
--- a/roles/satellite-clone/tasks/ensure_pulp_data_permissions.yml
+++ b/roles/satellite-clone/tasks/ensure_pulp_data_permissions.yml
@@ -1,0 +1,30 @@
+---
+- name: 'Ensure pulp group exists'
+  group:
+    name: pulp
+    system: true
+
+- name: 'Ensure pulp user exists'
+  user:
+    name: pulp
+    group: pulp
+    create_home: false
+    home: /var/lib/pulp
+    shell: /sbin/nologin
+    system: true
+
+- name: 'Check /var/lib/pulp ownership'
+  stat:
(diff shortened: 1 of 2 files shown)


roles/satellite-clone/tasks/ensure_pulp_data_permissions.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: 'Ensure pulp user exists'
```
ekohl, line 2: Does this also ensure the group exists? I don't know the Ansible `user` resource, but later on it's assumed the group exists. So either the description is incorrect or you're missing a step.

evgeni: It uses adduser and that by default does create the group.

But I could also create the group explicitly.

evgeni: (done)

roles/satellite-clone/tasks/ensure_pulp_data_permissions.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: 'Ensure pulp gorup exists'
```
ekohl, line 2:
```suggestion
- name: 'Ensure pulp group exists'
```

ekohl (approved):
Small typo, otherwise :+1:

PR #417, opened 2022-03-07
document 6.10 support

evgeni:
fixes: #404


evgeni:
I have decided to drop #402 from the "6.10 support milestone", as I've yet to hear back if my proposal really works, but additionally not having it doesn't block normal clone operations ("no-pulp" is really only for restoring customer backups for reproducers, and even then, they mostly work without the reset)

ekohl (approved):

PR #421, opened 2022-06-14
correct postgresql.conf path for el8

evgeni:


wbclark (approved):
this is needed for sat 6.11 on el8

PR #423, opened 2022-06-23
enable the satellite module for el8

evgeni:


ekohl (approved):
Modular metadata should **always** be there for all EL8 Satellite installs, right? You could consider making it a variable if you at some point want to support upstream cloning but that probably requires more work anyway.

evgeni:
Correct, it's always there.

And yeah, templating this is probably a good idea

ekohl (approved):

PR #432, opened 2023-04-11
replace bin/satellite-clone with a simple bash script

evgeni:
this avoids the weird "need to run with same python as ansible" depdendency


bin/satellite-clone
```diff
@@ -1,98 +1,38 @@
...
+
+options:
+  -h, --help        show this help message and exit
+  --step            interactive: confirm each task before running
+  --list-tasks      list tasks that will be run in the playbook
+  --flush-cache     clear the fact cache and run playbook
+  -y, --assume-yes  Assume yes for each question"
+}
+
+if [[ "$@" == "--help" ]] || [[ "$@" == "-h" ]]; then
```
ekohl, line 20: Does this work to check if an entry is in an array? If so, I learned something new.

evgeni: it doesn't :)

bin/satellite-clone
```diff
@@ -1,98 +1,38 @@
...
+export ANSIBLE_CONFIG
+
+usage() {
+  echo "usage: satellite-clone [-h] [--step] [--list-tasks] [--flush-cache] [-y]
+
+options:
+  -h, --help        show this help message and exit
+  --step            interactive: confirm each task before running
+  --list-tasks      list tasks that will be run in the playbook
+  --flush-cache     clear the fact cache and run playbook
```
evgeni, line 16: honestly, I am considering dropping those parameters alltogether.

evgeni:
@Griffin-Sullivan this will affect the check we added in https://github.com/SatelliteQE/robottelo/pull/10972, you'd now have to pass `--assume-yes --list-tasks` for it to work (and it's using an undocumented feature then, which I think is fine tho)

bin/satellite-clone
evgeni: :potato:

bin/satellite-clone
```diff
@@ -1,98 +1,41 @@
...
+
+if [[ "$1" == "--assume-yes" ]] || [[ "$1" == "-y" ]]; then
+  CONFIRM=true
+  shift
+else
+  CONFIRM=false
+fi
+
+if [[ "${CONFIRM}" != "true" ]]; then
+  read -p "This will initiate satellite-clone. Do you want to proceed? (y/n): " result
```
ekohl, line 35: You can capitalize N to indicate that it's the default if they press enter. And shellcheck suggests using `-r`, even though it won't really affect us here, it's still a good thing to have.
```suggestion
  read -r -p "This will initiate satellite-clone. Do you want to proceed? (y/N): " result
```

ekohl (approved):

PR #437, opened 2023-04-13
do not enable the profile_tasks callback

evgeni:
it's not available in ansible-core and noone complained for it being missing since we moved to EL8


ekohl:
Please rebase

evgeni:
Please, rebased.

ekohl (approved):

PR #439, opened 2023-06-09
report satellite version on miss-match

evgeni:


evgeni:
looks like this:

```
    TASK [satellite-clone : Check for satellite version in metadata file] **********
    fatal: [localhost]: FAILED! => {"changed": false, "msg": "Satellite version 6.11 is not supported or found. Only Satellite 6.12 or newer is supported."}
```

ekohl (approved):

PR #441, opened 2023-06-21
add a fallback for satellite vars

evgeni:
this allows adding support for newer satellite versions easier


ekohl (commented):
Does this mean it reads both or the first one that exists? I can see reading both with a specific overrides can make branching easier. If it only reads one then it means changing the defaults requires you to copy the defaults file to all supported versions and then keep them in sync.

roles/satellite-clone/tasks/main.yml
```diff
@@ -17,7 +17,14 @@
   include_vars: "el{{ ansible_distribution_major_version }}.yml"
 
 - name: Set Satellite version dependent variables
-  include_vars: "satellite_{{ satellite_version }}.yml"
+  include_vars: "{{ lookup('ansible.builtin.first_found', params) }}"
+  vars:
+    params:
+      files:
+        - "satellite_{{ satellite_version }}.yml"
```
ehelms, line 24: Does this then just exist for compatibility reasons?

evgeni:
Right now it  reads the first found. My hope is not to need those overrides at all at some point.

ekohl:
Perhaps it's fine to start with this and work out issues along the way. I was thinking along these same lines as a solution.

evgeni:
okay, how about now? :D

ehelms:
LGTM. Is this the goal for satellite-clone releases to be able to support multiple Satellite versions? If yes, is that to make it easier to backport versus having a version branch that correlates?

evgeni:
> LGTM. Is this the goal for satellite-clone releases to be able to support multiple Satellite versions? If yes, is that to make it easier to backport versus having a version branch that correlates?

I think with this change in, we can swap the version check logic to "6.12 or newer" and just let it sit until things break.

(In the past we *had* to support multiple versions due to the single-maint-repo, today it's just less hastle for me?)

ekohl (approved):
This is still a draft, but :+1: on the concept. Implementation also looks ok, but I'm not an ansible master so don't trust me too much on that.

evgeni:
I did test-run this and it executed as required, so going to merge it :)

PR #448, opened 2024-03-05
don't run installer again after we restored the backup

evgeni:
foreman-maintain runs the installer after the restore for us, no need to repeat things

saves about a minute in my tests :)


evgeni:
it's been doing this since https://github.com/theforeman/foreman_maintain/pull/553

ehelms (approved):

PR #450, opened 2024-03-06
verify interfaces based on backup metadata not answers

evgeni:


roles/satellite-clone/tasks/interface_check.yml
```diff
@@ -1,55 +1,11 @@
...
   when:
-    - dns_enabled.value
-    - dhcp_enabled.value
-    - dns_interface.value not in network_interfaces.msg
-    - dhcp_interface.value not in network_interfaces.msg
\ No newline at end of file
+    - proxy_config.dns
+    - proxy_config.dhcp
+    - proxy_config.dns_interface not in network_interfaces
+    - proxy_config.dhcp_interface not in network_interfaces
```
ekohl, line 9: DHCP and DNS are independent. So it's `(proxy_config.dns and proxy_config.dns_interface not in network_interfaces) or (proxy_config.dhcp and proxy_config.dhcp_interface not in network_interfaces)`.

ekohl (changes requested):

roles/satellite-clone/tasks/interface_check.yml
evgeni: Fair, even tho this is an old bug.

ekohl (approved):

PR #453, opened 2024-06-04
don't stop/start crond and don't start services after restore

evgeni:
foreman-maintain takes care of all that, no need to duplicate work

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -257,22 +257,8 @@
 
 - name: Finalize clone
   block:
-  # Cron services can interfere with restoring the backup
-  - name: stop crond
-    systemd:
-      name: crond
-      state: stopped
-
   - name: restore using foreman-maintain
     command: foreman-maintain restore --assumeyes {{ backup_dir }}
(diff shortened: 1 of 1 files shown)


roles/satellite-clone/tasks/main.yml
```diff
@@ -257,23 +257,9 @@
...
-    systemd:
-      name: crond
-      state: started
-
-  - name: Restart services
-    command: foreman-maintain service start
-
     # We don't use pause module here because it displays a prompt
     # to use ctrl+c which would break out of a production script.
   - name: Wait for foreman-tasks service to start
```
ekohl, line 265: Is the sleep still needed? By now, don't we guarantee services being available after `systemctl start $service`?

ekohl (approved):

roles/satellite-clone/tasks/main.yml
evgeni: one hunk at a time ;)

foreman-tasks takes a moment to assemble, so `hammer ping` often shows that one as "dead" if you do it *directly* after, but the sleep is certainly too long

PR #455, opened 2025-01-15
use runuser instead of become to check postgres perms

evgeni:
fapolicyd doesn't like when we try to execute Python as `postgres`

### DIFF
--- a/roles/satellite-clone/tasks/main.yml
+++ b/roles/satellite-clone/tasks/main.yml
@@ -155,9 +155,7 @@
     recurse: yes
 
 - name: test foreman.dump file is readable by postgres user
-  command: "test -r {{ backup_dir }}/foreman.dump"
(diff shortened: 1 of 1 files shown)


evgeni:
A `runuser` become method would not have helped. It's the whole become mechanism that breaks it for us.

`become` in Ansible works like this:
* somehow become (heh) the user that was requested
* write out the (generated) python code to disk
* execute the python code (still as that user)
* report the result back to the controller

Now the problem is that fapolicyd detects the python code (correctly) as python code, but the default policy is "don't let users run random python code".

But the policy does allow *root* to run random Python code.
So as long as we don't use `become_user: someone_not_root`, our code gets written and executed as root (and only then, inside python does shell out to `runuser` to switch to `someone_not_root`), which is fine as far as fapolicyd is concerned.

